PR #43, opened 2023-02-16
DSEGOG-230 Add record timestamp to channel sample

MRichards99:
As discussed with the frontend team, this change adds a timestamp (taken from record metadata where the data comes from) to the channel sample on the channel summary endpoint. This changes the samples from a list of values to a list of dictionaries, containing a timestamp and value. Example response:

```json
{
    "first_date": "2022-04-07T14:16:16",
    "most_recent_date": "2022-04-08T16:58:57",
    "recent_sample": [
        {"2022-04-08T14:16:16": -8582000.0},
        {"2022-04-09T14:16:16": -8461000.0},
        {"2022-04-10T14:16:16": -8535000.0}
    ]
}
```

### DIFF
--- a/operationsgateway_api/src/models.py
+++ b/operationsgateway_api/src/models.py
@@ -145,7 +145,7 @@ class ChannelManifestModel(BaseModel):
 class ChannelSummaryModel(BaseModel):
     first_date: datetime
     most_recent_date: datetime
-    recent_sample: List[Union[int, float, str]]
+    recent_sample: List[Dict[datetime, Union[int, float, str]]]
 
     class Config:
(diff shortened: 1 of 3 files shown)


MRichards99:
I had a couple of issues getting tests to pass (due to MD5 checksums) after I merged main in. I had to re-ingest the data and that worked. Exact command that I used:
`poetry run python util/ingest_hdf.py -p ../operationsgateway-test-data/dev_server -w -d -i ../og-images/ -U backend -P back`

Just wanted to let you know in case you come across the same thing when you review this

PR #45, opened 2023-02-23
DSEGOG-226 Add comments to example config file

MRichards99:
A small PR as part of the 'improvements to codebase' epic, just so the config file makes a bit more sense for anyone who's not us.


operationsgateway_api/config.yml.example
```diff
@@ -20,6 +24,7 @@ mongodb:
   max_documents: 1000
   image_store_directory: /home/user/og-images
 auth:
+  # Create a private/public key pair for the auth functionality
```
kevinphippsstfc, line 27: This should probably just say "Path to a private/public key pair for the auth functionality". 
For me (and possibly only me!) it is ambiguous whether "Create" is telling the user to create the keys or whether the application will be creating the keys at the specified location. Maybe a second line of comment is needed to clarify that the keys need to be created in advance and put at the specified location?

kevinphippsstfc (changes requested):

PR #47, opened 2023-02-23
DSEGOG-218 Change logger config to use a rotating log handler

MRichards99:
This PR changes the logging config to use a rotating log file hander, so a new log file is created when the current one reached 5MB.

I had a go at doing this originally but I couldn't get to a config that I was happy with - I was being a bit picky with having startup output into the log file and into stdout (so you know that the startup of the API has completed without having to look in the log file). I've managed to sort this and I'm quite happy with these changes:
- I've tested that the rotation works. I changed `maxBytes` to 10000 and verified it works
- The startup output (where FastAPI tells you processes are starting etc.) is in stdout and the log file
- Print statements still print to stdout for a quick bit of debugging

### DIFF
--- a/operationsgateway_api/logging.ini.example
+++ b/operationsgateway_api/logging.ini.example
@@ -22,7 +22,7 @@ propagate=0
 [handler_fileHandler]
 class=logging.handlers.TimedRotatingFileHandler
 formatter=fileFormatter
-args=('logs.log', 'D', 1, 20,)
+args=('operationsgateway-api.log', 'D', 1, 20,)
 
 [handler_consoleHandler]
 class=StreamHandler


kevinphippsstfc (changes requested):
I hadn't really looked too closely at the logging config before but now that I see how it is done I think it needs changing. Currently we have a bit of a mix - you can set up 3 of the logging parameters in the config.yml file and the others (such as the logging handler class being changed here) are hard coded and can't be changed.

What I think we should move to is a logging.ini file where all of the logging settings are stored and remove the logging settings in the config.yml file. That way the logging is totally configurable via the ini file and nothing is hard coded. An example of this is the logging.ini file in the pollcat repo at: https://github.com/ral-facilities/pollcat/blob/master/logging.ini

By the way, that logger uses a TimedRotatingFileHandler which creates a new log file for each day which I find is particularly convenient on production systems.

MRichards99:
I've had a look this afternoon and I'm struggling to replicate the same behaviour that we currently have. I've come to the following `logging.ini` file, a lot of which is based off the Pollcat example:

```
[loggers]
keys=uvicorn.access,root

[handlers]
keys=fileHandler

[formatters]
keys=fileFormatter

[logger_root]
level=DEBUG
handlers=fileHandler
qualname=root
propagate=0

[logger_uvicorn.access]
handlers=fileHandler
qualname=uvicorn.access
propagate=0

[handler_fileHandler]
class=logging.handlers.TimedRotatingFileHandler
level=DEBUG
formatter=fileFormatter
args=('logs.log', 'D', 1, 20,)

[formatter_fileFormatter]
format=[%(asctime)s] %(levelname)s - %(message)s
datefmt=
```

It _kind of_ works but when you start the API, you used to see the logging in the console to show that it was starting/successfully started (the same for when you turn the API off). However, I cannot replicate this behaviour when using `logging.ini` - it doesn't even go in the log file itself which I'm surprised about. I rely on this behaviour to know if the API is happy on startup or if I've missed a config option or something - otherwise you don't really know how the startup went.

I'm a bit stuck with it really, I got the behaviour just right with the PR (albeit needing changing to the timed handler) but I can't seem to get it right with `logging.ini`. Not really sure why, the logging config looks close to what the config is in the PR so I thought it would work but I've tried all sorts of permutations

MRichards99:
After taking another look, I've been successful in using a specific `logging.ini` file to configure the logging. I needed to set up a specific console handler, something that wasn't required when it was configured via a dictionary. You now get the logging from Uvicorn and the API in the console (see below for an example output using INFO level logging). I've added a `logging.ini.example` which follows the typical ICAT example file procedure (added a quick note about this in `README.md`). 

Example console logging:
```bash
INFO:    Logging now setup
INFO:    Will watch for changes in these directories: ['/root/seg_dev/operationsgateway-api']
INFO:    Uvicorn running on http://127.0.0.1:8000 (Press CTRL+C to quit)
INFO:    Started reloader process [16179] using StatReload
INFO:    Logging now setup
INFO:    Logging now setup
INFO:    Getting records by query
INFO:    Sending find() to MongoDB, collection: records
INFO:    Getting query results and converting them into a list: 1000
INFO:    127.0.0.1:52768 - "GET /records?limit=2 HTTP/1.1" 200
INFO:    Getting channel metadata from database
INFO:    Sending find_one() to MongoDB, collection: channels
INFO:    127.0.0.1:52768 - "GET /channels HTTP/1.1" 200
```

I've made the required changes on Ansible (see https://github.com/ral-facilities/operationsgateway-ansible/pull/7) so this PR can be deployed.

This PR is ready to be re-reviewed.

operationsgateway_api/logging.ini.example
```diff
@@ -0,0 +1,38 @@
...
+[logger_uvicorn.access]
+level=INFO
+handlers=fileHandler,consoleHandler
+qualname=uvicorn.access
+propagate=0
+
+[handler_fileHandler]
+class=logging.handlers.TimedRotatingFileHandler
+formatter=fileFormatter
+args=('logs.log', 'D', 1, 20,)
```
kevinphippsstfc, line 25: Why is the log file called logs.log? Can it be called operationsgateway-api.log like it seems to be on my development setup? Or is this following some kind of convention?

kevinphippsstfc (commented):
This looks good - just a question I raised about the log file name.

It's great that the setup is now in a config file rather than in code, but I'm always surprised how complicated logging setups can seem when normally they are just sending log messages to either the console or a file!

PR #50, opened 2023-03-01
Adding badges to README

agbeltran:
adding CI and codecov badges - at the moment, the latter one shows unknown but it seems to be a problem on their side


agbeltran:
Good point @kevinphippsstfc! yes, it would be good to add the license badge too - I will update this PR with a new commit

MRichards99 (approved):

PR #51, opened 2023-03-15
DSEGOG-179 DSEGOG-180 DSEGOG-181 DSEGOG-238 Experiment IDs Functionality

MRichards99:
This PR adds the functionality to retrieve experiments from the Scheduler, store them in MongoDB and allow the frontend to get a list of them. The API has two new endpoints for this functionality:
- `POST /experiments` - query the Scheduler for new experiments and store them in MongoDB
- `GET /experiments` - return a list of experiments from MongoDB

While the Scheduler stores lots of data about an experiment, we're only interested (and therefore store) the following things:
- Experiment ID
- Start & end dates of experiment
- Part number of the experiment
- `_id` is the experiment ID and part number joined by a dash

We get this data using two of the Scheduler's calls, `getExperimentDatesForInstrument` and `getExperiments`. These calls along with `login` (and client creation) are inside a specific class, `SchedulerInterface` as I wanted to separate that from the general experiment IDs functionality. We're using `zeep` as the SOAP client rather than `suds-community` that we use for Python ICAT. This is because I experienced various HTTP issues when using suds, which I didn't after I switched to Zeep. If we need to switch back, it's as simple as changing the import and editing the error handling decorator which deals with SOAP error messages.

The duplicate part selector is a piece of functionality that deals with experiments that contain multiple parts with the same part number. This is a bug on the Scheduler but it could be a while until it's fixed so I've implemented a workaround in the form of this part selector; it looks at the part status and chooses the most appropriate part to use from all the duplicates. I didn't see any duplicate parts on the production Scheduler data I've looked at but it was present on the dev system. It's definitely a bit of an edge case but if it's not fixed on the Scheduler before OperationsGateway goes into production, duplicate parts could be a thing (as part of setting up a new facility).

Retrieving experiments from the Scheduler can happen in two ways, by using the appropriate endpoint or via a background runner. The background runners use a cron format to configure and mean that new experiments can be retrieved once every defined time period (I imagine once a week, once a month etc.). If the Scheduler calls fail, it waits a configured amount of time before retrying. If it fails a second time, an error will be logged and the task will wait until the next time it should run (e.g. next week). There's a TODO and a Jira issue (DSEGOG - 228) to the response to a second a failure more effective in the future as it becomes clearer how monitoring of OperationsGateway will be done.

The `Experiment` class itself stores and retrieves experiments to/from MongoDB (specifically the `experiments` collection). It puts data from the Scheduler into an `ExperimentModel` (new Pydantic model), before putting it into the database. I'd follow the code (and what it calls) in `store_experiments_from_scheduler()` to get a general understanding of what goes on - functions are documented and type hinted where possible.

As well as the experiments data, the `experiments` database collection has a record containing the last time the collection was updated (find `collection_last_updated` in `Experiment`). This is updated every time new experiments are stored in the database so that the start date is appropriate for the next time the Scheduler is contacted. This is done because each time the Scheduler is contacted, we're only getting the experiments that aren't currently stored in MongoDB (rather than the entire list).

The tests provide fairly good coverage of the new functionality (94% on `operationsgateway_api/src/experiments/`) as there's an entire directory inside of `test/` dedicated to testing this functionality. There's one file per class which don't require any access to the Scheduler (this has all been mocked) so they'll run on GitHub Actions. The only test that requires the Scheduler is the integration test for `GET /experiments` - I have added a skip on the test if it's running on GitHub Actions.


kevinphippsstfc (changes requested):
A couple of issues that we need to address before merging this in:

1. I think we should have an option in the config file to switch the functionality off
2. Currently one of these timed tasks runs for each of the instances of the OG-api code (workers) running on the server, which means that multiple tasks will run at the same time trying to do exactly the same thing (probably with mixed results depending on exact timings). We either need to find a way to ensure that only one instance of the timed task can run, or perhaps we need to extract the code for this into a separate service. This would not be so easy as it shares common config, database etc code with the rest of the application, and two applications would need to be configured and set up rather than just one.

operationsgateway_api/src/experiments/unique_worker.py
```diff
@@ -0,0 +1,133 @@
...
+
+log = logging.getLogger()
+
+
+class UniqueWorker:
+    """
+    Where multiple workers are launched for this API (by uvicorn for example), this
+    class is used to assign a task to a single worker. For example, the background task
+    to contact the Scheduler on a regular basis only needs to be performed by a single
+    worker. A process ID is to the 'assigned worker' is written to a file and checked
```
kevinphippsstfc, line 18: "A process ID is to the 'assigned worker'" needs a bit of rewording

test/experiments/test_unique_worker.py
```diff
@@ -0,0 +1,236 @@
...
+import pytest
+
+from operationsgateway_api.src.experiments.unique_worker import (
+    assign_event_to_single_worker,
+    UniqueWorker,
+)
+
+
+class TestUniqueWorker:
+    # TODO - path doesn't seem to be mocking correctly
```
kevinphippsstfc, line 12: Did you mean to leave this TODO in?

kevinphippsstfc (changes requested):

test/experiments/test_unique_worker.py
MRichards99: I didn't, but I'm glad I did (and you spotted it)! The mocking was being ignored, but I've fixed this by mocking directly onto the class variable. Fixed in c861b49bbf35e931dd8b849155181d38ca7a320a

operationsgateway_api/src/experiments/unique_worker.py
MRichards99: Fixed in c861b49bbf35e931dd8b849155181d38ca7a320a, hopefully it makes more sense now?

kevinphippsstfc (approved):

PR #52, opened 2023-03-16
Add API startup output to ingestion script

MRichards99:
A small PR to improve the ingestion script. The other day, the ingestion script get stuck in a loop on the CI due to missing config. I didn't know this was the problem without SSH'ing into the runner so I've made the script able to output the API startup output. 

When the API fails to startup, the script will also now exit, instead of being stuck in a loop of searching for something on the specified port


MRichards99:
The test failures are because of the typo fix on the channel name (the x y one) which is fixed in #51. The dependency safety failure is also fixed in that PR

MRichards99:
I've merged #47 into this branch and made a couple of changes to make this PR work with the logging config changes

PR #53, opened 2023-03-16
DSEGOG-217 Fix linting errors

MRichards99:
A small PR to fix the `# noqa` comments that we have. A lot them can be ignored (which I've done in `.flake8`) due to the inner workings of FastAPI, as recommended by some GitHub issue discussions.

There were some genuine code issues in `mongo/interface.py` with mutable data types in the function definition. I've fixed these by setting the defaults to `None`, checking for that in the function and setting them to empty dicts


MRichards99:
The test failures are because of the typo fix on the channel name (the x y one) which is fixed in https://github.com/ral-facilities/operationsgateway-api/pull/51. The dependency safety failure is also fixed in that PR

PR #61, opened 2023-05-31
DSEGOG - 251 Organise colourmaps into different categories

MRichards99:
This PR organises colourmaps into several categories, as listed in https://matplotlib.org/stable/tutorials/colors/colormaps.html. This will allow the frontend to display some categories in the main dropdown list, and the rest in an extended list (accessible by a checkbox). I've created a mapping file in `colourmap_mapping.json`.

This is stored in `operationsgateway_api/src/records/` as it won't be used as a config file - the only time that colourmaps will be added/removed from Matplotlib is when we upgrade the version of the dependency. The frontend team has asked us to provide a 'master' list which they'll manipulate on their end to get two lists of colourmaps as users request. So if a particular colourmap is removed from the frontend, the frontend will remove it on their end, but it's still accessible if someone queries the API directly.

I've written tests for the code I've written, including tests to check all colourmaps listed in the mapping file contains valid colourmaps and the mapping file contains every single colourmap available in Matplotlib, thus allowing us to easily monitor any additional colourmaps that are added in Matplotlib in the future.


test/images/test_colournap_mapping.py
```diff
@@ -0,0 +1,74 @@
...
+            colourmap_name,
+        )
+        assert available_map == expected_availability
+
+    def test_mapping_file_valid(self):
+        colourmap_mapping = self.load_mapping_file()
+        flat_colourmap_list = [
+            colourmap for group in colourmap_mapping.values() for colourmap in group
+        ]
+        mpl_colourmaps = plt.colormaps()
```
kevinphippsstfc, line 59: A very minor point but as you are only using mpl_colourmaps once below I don't think it is worth assigning it to a variable. Just use plt.colormaps() directly below.

test/images/test_colournap_mapping.py
```diff
@@ -0,0 +1,74 @@
...
+
+        # Checking the all colourmaps listed in the mapping file are valid colourmaps
+        for colourmap in flat_colourmap_list:
+            assert colourmap in mpl_colourmaps
+
+    def test_mapping_file_contains_all_colourmaps(self):
+        colourmap_mapping = self.load_mapping_file()
+        mpl_colourmaps = plt.colormaps()
```
kevinphippsstfc, line 67: Same comment as for this variable in previous test

test/images/test_colournap_mapping.py
```diff
@@ -0,0 +1,74 @@
...
+        assert available_map == expected_availability
+
+    def test_mapping_file_valid(self):
+        colourmap_mapping = self.load_mapping_file()
+        flat_colourmap_list = [
+            colourmap for group in colourmap_mapping.values() for colourmap in group
+        ]
+        mpl_colourmaps = plt.colormaps()
+
+        # Checking the all colourmaps listed in the mapping file are valid colourmaps
```
kevinphippsstfc, line 61: Remove "the"

kevinphippsstfc (changes requested):

test/images/test_colournap_mapping.py
MRichards99: Fixed in d4862f16ad104b69e0fa63d24b901b9ad176ef7c

test/images/test_colournap_mapping.py
MRichards99: Fixed in d4862f16ad104b69e0fa63d24b901b9ad176ef7c

test/images/test_colournap_mapping.py
MRichards99: My thinking here was to prevent calling `plt.colormaps()` multiple times but I've just tested it both ways and it makes absolutely no difference to the speed of the test (it takes 0.01 seconds). I've made the change in d4862f16ad104b69e0fa63d24b901b9ad176ef7c.

PR #66, opened 2023-07-31
DSEGOG-255 EPAC Simulated Data

MRichards99:
This PR contains a number of scripts in `util/realistic_data/` as a mechanism to generate simulated data and ingest it onto an instance of the API. I've written some documentation about this in `util/realistic_data/README.md` so I won't go into much detail about it in this PR, but direct you to that file.

There are also a small number of changes made to the source code of the API to support the change in how timestamps will be stored in HDF files.

https://github.com/ral-facilities/operationsgateway-ansible/pull/10 shows how this will be setup and used for the dev server.


MRichards99:
The test coverage in this PR is down because there are no specific tests for the HDF ingestion code, which is where the modifications to the source code are. Currently, this code is effectively tested with the ingestion script. My suggestion would be that improving test coverage on the ingestion code would be a good task for Will to do (i.e. complete DSEGOG - 262)?

MRichards99:
Once #77 is merged, I'll have to make changes how experiments are assembled (`calendar_conversion.py` and the JSON file that is an output of that script). Check the config for simulated data too

util/realistic_data/config.yml.example
```diff
@@ -0,0 +1,30 @@
...
+database:
+  hostname: localhost
+  port: 27017
+  name: opsgateway
+  remote_experiments_file_path: /tmp/experiments_for_mongoimport.json
+echo:
+  endpoint_url: https://s3.echo.stfc.ac.uk
+  access_key: access_key
+  secret_key: secret_key
+  simulated_data_bucket: og-realistic-data
```
Will-Cross1, line 17: og-ci-simulated-data should work better for the users

Will-Cross1 (approved):
**Looks good to me**!

just need to resolve the merge conflicts 
and decide wether to change the simulated_data_bucket in the example config to og-ci-simulated-data

PR #67, opened 2023-08-03
DSEGOG-261 Use ObjectIds in ExperimentModel

MRichards99:
This PR changes the `_id` in the `ExperimentModel` to use `ObjectId`s. Previously, this field was filled with a concatenation of the experiment ID and the part number. However, in the simulated EPAC data, experiment IDs for maintenance periods are stored as `maintenance`. In the case of the generated data, they were all experiments with a part number of 1.

Given current discussions regarding parts, it's unclear if this'll be the case in production so I've decided to move to using ObjectIds. We use these in the user session functionality so wasn't too bad to make this change (I was a bit scared of them before). Even the tests didn't require too much work!

### DIFF
--- a/operationsgateway_api/src/experiments/experiment.py
+++ b/operationsgateway_api/src/experiments/experiment.py
@@ -72,15 +72,39 @@ class Experiment:
         that haven't yet been inserted into the database
         """
 
+        inserted_ids = []
         for experiment in self.experiments:
-            await MongoDBInterface.update_one(
-                "experiments",
-                {"_id": experiment.id_},
-                {"$set": experiment.dict(by_alias=True)},
-                upsert=True,
+            experiment_data = experiment.dict(
+                by_alias=True,
+                exclude_unset=True,
+                exclude={"id_"},
             )
 
+            update_result = await MongoDBInterface.update_one(
+                "experiments",
+                experiment_data,
+                {"$set": experiment_data},
+                upsert=True,
+            )
+            # This means the document was updated, not inserted. The _id needs to be
+            # found to be used as a response
+            if update_result.upserted_id is None:
+                updated_experiment = await MongoDBInterface.find_one(
+                    "experiments",
+                    experiment_data,
+                )
+
+                log.debug(
+                    "Experiment was updated, found experiment _id: %s",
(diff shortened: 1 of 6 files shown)


operationsgateway_api/src/experiments/experiment.py
```diff
@@ -72,16 +72,40 @@ async def store_experiments(self) -> None:
...
+                    "Experiment was updated, found experiment _id: %s",
+                    str(updated_experiment["_id"]),
+                )
+                inserted_ids.append(f"Updated {str(updated_experiment['_id'])}")
+            else:
+                inserted_ids.append(str(update_result.upserted_id))
 
         await self._update_modification_time()
 
+        return inserted_ids
```
kevinphippsstfc, line 107: The return type of the function needs updating now that something is being returned

test/experiments/test_experiment.py
```diff
@@ -69,6 +70,54 @@ async def test_get_experiments_from_scheduler(self, _, __, ___, ____, _____):
         )
         assert test_experiment.experiments == expected_experiments
 
+    @patch(
+        "operationsgateway_api.src.experiments.scheduler_interface.SchedulerInterface.__init__",
+        return_value=None,
+    )
+    @pytest.mark.asyncio
+    async def test_store_experiments(self, _):
```
kevinphippsstfc, line 78: Test could do with a few more comments to explain what is happening.

kevinphippsstfc (changes requested):

operationsgateway_api/src/experiments/experiment.py
MRichards99: Fixed in 592a500182053d2aff91fdaabb6d09d92ddc6af3

test/experiments/test_experiment.py
MRichards99: Fixed in 592a500182053d2aff91fdaabb6d09d92ddc6af3

MRichards99:
I've made the suggested changes and the CI has passed. I've marked this PR for re-review

kevinphippsstfc (approved):

PR #68, opened 2023-08-17
DSEGOG-263 Upgrade `certifi` for security vulnerability

MRichards99:
As per a recent Dependabot alert (see the security tab of this repo if you haven't seen it yet), the `certifi` dependency needs upgrading. A quick `poetry update certifi` did the trick, giving us the patched version.

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -412,34 +412,34 @@ test = ["unittest"]
 
 [[package]]
 name = "cryptography"
-version = "41.0.2"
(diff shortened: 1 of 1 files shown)


MRichards99:
I've also just upgraded `cryptography` as I realised that also needs upgrading according to Dependabot. We went from 41.0.2 to 41.0.3.

kevinphippsstfc (approved):

PR #70, opened 2023-08-23
DSEGOG-258 Multiple Instrument Support for Experiments Functionality

MRichards99:
This PR allows the API to contact the Scheduler and ask for details on experiments that take place on multiple instruments where previously, only data about one instrument could be retrieved. In the config, the instruments that the API contacts are configured using by specifying the instrument names in list. In the example we currently have a single instrument but I expect that to be two in production (EA1 & EA2). 

`get_experiments_from_scheduler()` is a good place to start looking at the changes - the code now iterates through the list of configured instrument names and sends a call per instrument to the Scheduler for experiments on that instrument within the relevant start and end dates.

`_map_experiments_to_part_numbers()` now returns a list of `ExperimentPartMappingModel`, a new Pydantic model added in this PR. This replaces a list of dictionaries; the change was partly to improve code readability (as Pydantic models are easier to read/debug than dictionaries) but also because this data structure needs to keep track of which instrument the experiment is linked with. The instrument name from this data structure is used in `_generate_id_instrument_name_pairs()` - before the instrument name was just grabbed from the config but obviously this is no longer possible as there's multiple choices.

In `_generate_id_instrument_name_pairs()`, I was facing an issue where the list of experiment ID & instrument name pairs included duplicates. Converting the list of dictionaries to a set of dictionaries seems like the obvious thing to do to remove duplicates, but dictionaries aren't hashable so you can't do that straight away. Stack Overflow suggested a list comprehension which puts the data from the dictionaries into tuples (which are hashable), convert the list to a set and convert the tuples back into dictionaries. I have added comments explaining this so hopefully it's clear to someone reading the code for the first time?

A list of `ExperimentPartMappingModel`s are passed into `_extract_experiment_data()` (this used to be a list of dictionaries). As lookups cannot be performed on the model (but they could be done with dictionaries), I've added a new function, `_get_mapping_model_by_experiment_id()`. This takes a list of mapping models and an experiment ID, searching through the list until the data is found with a matching experiment ID. 

I've edited existing unit tests to cope with this change and added a new test (`test_valid_non_selected_extract_experiment_data()`) that covers a situation where `_extract_experiment_data()` might be passed some experiment parts which haven't been selected. There are also two new tests for the new `_get_mapping_model_by_experiment_id()` function.

### DIFF
--- a/operationsgateway_api/src/experiments/experiment.py
+++ b/operationsgateway_api/src/experiments/experiment.py
@@ -90,7 +90,7 @@ class Experiment:
         self,
         # List of zeep experimentDateDTO's
         experiments: List[Any],
-        instrument_name: str
+        instrument_name: str,
     ) -> List[ExperimentPartMappingModel]:
         """
         Given a list of experiments retrieved from the Scheduler, extract RB numbers
--- a/operationsgateway_api/src/experiments/experiment.py
+++ b/operationsgateway_api/src/experiments/experiment.py
@@ -230,20 +230,16 @@ class Experiment:
         a matching experiment ID
         """
 
-        part_mapping = None
         for mapping in mapping_models:
             if mapping.experiment_id == experiment_id:
                 return mapping
 
-        if not part_mapping:
-            log.error(
-                "Part mapping for %s failed. Mapping to search through: %s",
-                experiment_id,
-                mapping_models,
-            )
-            raise ExperimentDetailsError(
-                f"Lookup of part mapping {experiment_id} failed",
-            )
+        log.error(
+            "Part mapping for %s failed. Mapping to search through: %s",
+            experiment_id,
+            mapping_models,
+        )
+        raise ExperimentDetailsError(f"Lookup of part mapping {experiment_id} failed")
 
     def _detect_duplicate_parts(
         self,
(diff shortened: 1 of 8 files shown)


operationsgateway_api/src/experiments/experiment.py
```diff
@@ -169,6 +197,31 @@ def _extract_experiment_data(
...
+        self,
+        experiment_id: int,
+        mapping_models: List[ExperimentPartMappingModel],
+    ):
+        """
+        Search through a list of experiment part mapping models and return the one with
+        a matching experiment ID
+        """
+
+        part_mapping = None
```
kevinphippsstfc, line 210: Setting this variable and then checking for it on line 215 are not required. They can be removed and then the indentation of the remaining code adjusted without affecting any functionality.

kevinphippsstfc (changes requested):

operationsgateway_api/src/experiments/experiment.py
MRichards99: Fixed in 39a16f708d7b3579f201f6bc093f0172465699c8

MRichards99:
I've made the suggested changes and the CI passes. I've marked this PR for re-review

kevinphippsstfc (approved):

PR #74, opened 2023-10-03
DSEGOG-270 Fix cleanup of Echo buckets

MRichards99:
I noticed that when GitHub Actions fails, the bucket that's created at the start of the job isn't cleaned up. This was happening because the steps . This PR fixes that bug by adding an extra condition to the if statement so buckets are removed if a step fails. You can see this works on the following run where I forced a failure in the CI job - https://github.com/ral-facilities/operationsgateway-api/actions/runs/6394539896/job/17356143894


MRichards99:
@Will-Cross1 I've added you as a reviewer for this pull request, can you review it for me please? Happy to explain the changes and any context behind the changes should you have any questions

PR #75, opened 2023-10-05
DSEGOG-273 added a user and set up poetry to work with Macs

Will-Cross1:
added Will as a user and set up the poetry requirements

changed the requirements to work with a Mac using python 3.8


MRichards99 (approved):
Looks good, ready to merge.

PR #80, opened 2023-11-09
DSEGOG-275 improve image tests

Will-Cross1:
changed the old md5 hashing to perceptual hashing

this was tested and perceptual hashing managed to keep the same hash after dependancies updated and the md5 hashes all changed


MRichards99:
@Will-Cross1 looks like GitHub Actions is failing on the tests. Take a look of the details of the run and look through the logs to see what's going on - I think there's something going on with the dependencies. Once it's fixed, I'll do a review :)

test/endpoints/test_get_record_by_id.py
```diff
@@ -52,19 +52,19 @@ def test_valid_get_record_by_id(
             pytest.param(
                 "20220408132830",
                 {
-                    "N_COMP_FF_IMAGE": "2fa3239b5f46b110964f02a2e6c6dd62",
-                    "N_COMP_NF_IMAGE": "16624db9b93bfe8176cd275d24522bc5",
-                    "N_LEG1_GREEN_NF_IMAGE": "1fcf716b4f670dc4a901ca77ea0b8ee5",
+                    "N_COMP_FF_IMAGE": "8c997363cc8c3373",
+                    "N_COMP_NF_IMAGE": "c2c63f8f783c3138",
+                    "N_LEG1_GREEN_NF_IMAGE": "8785789c6a7a6761",
```
MRichards99, line 57: The parameterised variable for this is called `expected_thumbnail_md5s`. Can you change it something like `expected_thumbnails_hashes` as we no longer use MD5 checksums please? `expected_thumbnail_md5s` is used across multiple files so you can do a search and replace in VSCode to change all instances in all files.

There's a few other instances of MD5 still being mentioned (e.g. in variable names and docstrings). Could you search for `md5` across the codebase and replace mention of MD5 with the type of hashing we use now please?

test/records/test_image.py
```diff
@@ -46,8 +47,10 @@ def test_create_thumbnail(self):
         )
         test_image.create_thumbnail()
 
-        thumbnail_checksum = hashlib.md5(test_image.thumbnail).hexdigest()
-        assert thumbnail_checksum == "d4f018fe6ab61b1706244e95b1d9bb96"
+        bytes_thumbnail = base64.b64decode(test_image.thumbnail)
+        img = PILImage.open(BytesIO(bytes_thumbnail))
+        thumbnail_checksum = str(imagehash.phash(img))
+        assert thumbnail_checksum == "0000000000000000"
```
MRichards99, line 53: Can you add a comment above this line to explain why we're asserting for a hash which contains all 0's please? I remember this is something that you asked about and it took me a few moments to work out why so a comment would likely be helpful to people in the future :)

pyproject.toml
```diff
@@ -6,12 +6,14 @@ readme = "README.md"
...
 [tool.poetry.dependencies]
-python = ">=3.8,<4.0"
+python = ">=3.8,<3.12"
 fastapi = "^0.103.2"
 motor = "^2.5.1"
 uvicorn = "^0.20.0"
 python-multipart = "^0.0.5"
 PyYAML = "^6.0"
+ImageHash = "4.3.1"
+scipy = "1.10.0"
```
MRichards99, line 16: Could you add a comment above the `scipy` line to explain that this is a dependency of `ImageHash` but is listed in our `pyproject.toml` to mitigate against a vulnerability in `scipy` version 1.9.3?

MRichards99 (changes requested):
Looks good, your work to test that the hashes don't change when dependency updates have show MD5 checksums changed before gives me confidence that this will work. Just a couple of minor things to change that I've noticed

MRichards99 (approved):
Brilliant, thanks for making the changes. Ready to merge when you are

PR #87, opened 2024-01-31
Add Python3.11 to tests, README detail and bump dependencies

patrick-austin:
## Changes
- Add detail on devel packages to README. Got the list from the Ansible repo, but since that's private (and possible to overlook) thought it was worth including a public facing reference.
- Update README and nox to target up to 3.11, and noted (at least one) reason why we can't target 3.12
- Included 3.10 and 3.11 in the CI matrix
- Bumped `motor` and `orjson`


MRichards99 (commented):
I just wanted to add a comment to remind me what I've seen when I come back to give a final review. Changes so far look good, just needs 3.10 & 3.11 CI sorting and the merge conflicts resolving

MRichards99 (changes requested):
To fix the issue with calling `login()` multiple places, just increase the number of minutes the access token is valid for. For GitHub Actions, you can do that at
https://github.com/ral-facilities/operationsgateway-api/blob/DSEGOG-287-Issues-setting-up-dev-environment/.github/ci_config.yml#L26

Looks good otherwise :)

patrick-austin:
> To fix the issue with calling `login()` multiple places, just increase the number of minutes the access token is valid for. For GitHub Actions, you can do that at https://github.com/ral-facilities/operationsgateway-api/blob/DSEGOG-287-Issues-setting-up-dev-environment/.github/ci_config.yml#L26
> 
> Looks good otherwise :)

Thanks - nice that this is configurable. I went looking and saw:
https://github.com/ral-facilities/operationsgateway-api/blob/878c44060cd56cdc0f7c1d2585a816d658e3b887/operationsgateway_api/src/routes/auth.py#L80
And (incorrectly) jumped to the conclusion the lifetime of the token was hardcoded so would have to just spam login or refresh requests. Though thinking about it for any significant amount of time should've seen that `604800` is a week in seconds so can't have been anything to do with a token expiring after an hour...

MRichards99 (approved):

PR #95, opened 2024-02-22
Indirect dependency lxml 5.1.0 causes an error

Will-Cross1:
symbol not found in flat namespace '_exsltDateXpathCtxtRegister'


MRichards99 (approved):
Looks good, merge when you're ready

PR #99, opened 2024-02-29
DSEGOG-302 Update Postman collection to use simulated CI data

MRichards99:
The Postman collection still references Gemini data (through channel names, IDs etc.) and now the main branch contains the simulated data stuff, we should move to use this in the Postman collection. This PR makes all relevant changes to make that move.

To test this PR, export the Postman collection to your Postman application and test each of the endpoints with the query parameters and ensure they work. There's one query parameter that won't work in `GET /records`, specifically `{"_id": {"$eq": "20230604000000"}}`. This is due to a bug in the API. An issue has been created for this in Jira (DSEGOG - 303) so it's fine that it doesn't work, it'll be resolved when that issue is fixed.

If you haven't moved to simulated data yet, please use the following guide to get that data in your development environment: https://github.com/ral-facilities/operationsgateway-api/blob/test-data-docs/docs/test_data.md

### DIFF
--- a/util/og_api_postman_collection.json
+++ b/util/og_api_postman_collection.json
@@ -1,9 +1,9 @@
 {
 	"info": {
-		"_postman_id": "d1e1b6a7-9aef-4d9d-82cb-3042fd13272e",
+		"_postman_id": "28dd0246-c384-4402-a74c-19993baa9a1f",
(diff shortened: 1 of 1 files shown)


Will-Cross1 (approved):
Looks good

PR #100, opened 2024-03-04
DSEGOG-285 Move waveforms to be stored in Echo

MRichards99:
To alleviate database issues, this pull request makes the change to store waveforms in Echo, instead of MongoDB. To do this, the data of a waveform is put into a JSON file and that's stored in Echo. As they're no longer stored in the database, they no longer need an idea; `waveform_id` has changed to `waveform_path`.

In an Echo bucket, there's now two main 'directories' - `images/` and `waveforms/`. The `Image` and `Waveform` classes have been adapted to have `get_relative_path()` and `get_full_path()` functions, whereby the latter gets the path including the main 'directory'. Hopefully comments and docstrings in the code make the distinction between the two clear.

In the code and the documentation, there was reference to Echo being used to store images only - I (think) I've changed that so those references are more generic or reference storage of both images and waveforms. This includes the config, moving some Echo stuff out of the images section and into a general `echo` section.

A big part of this PR has been the CI. As more data is being stored in Echo, the CI got slower. However, I was experiencing big slowdowns when using Echo with Actions - upon debugging, it appeared GitHub Actions would sit idle for 60 seconds at a time in some cases (at times it would be 2-3 minutes before idling stopped). My guess is that this is some rate limiting protection but it meant the ingestion script would take 2-3 hours, sometimes longer. To solve this, the API instance used for the ingestion script is plugged into a local setup of `minio` (a local S3 storage). The CI now runs in 13-15 minutes 🎉 

I also noticed that the use of `ThreadPool` in the ingestion script didn't seem to be helping so I've removed `download_and_ingest()` and HDF files are downloaded & ingested serially, looping through each filename.


Will-Cross1:
in the confog.yml.example it is missing upload_image_threads causing local pytests to fail

Will-Cross1 (changes requested):
In the confog.yml.example it is missing upload_image_threads causing local pytests to fail

operationsgateway_api/config.yml.example
```diff
@@ -12,13 +12,13 @@ images:
   # Colour maps for Matplotlib can be used
   default_colour_map: viridis
   colourbar_height_pixels: 16
-  echo_url: https://s3.echo.stfc.ac.uk
-  echo_username: username
-  echo_access_key: access_key
-  echo_secret_key: secret_key
-  image_bucket_name: test-bucket
-  upload_image_threads: 4
```
Will-Cross1, line 20: As per the requested change upload_image_threads should still exist in the example config

operationsgateway_api/src/records/waveform.py
```diff
@@ -1,34 +1,47 @@
...
+    def to_json(self):
+        """
+        Use `self.waveform` and return a JSON file stored in a BytesIO object
+        """
+        b = BytesIO()
+        b.write(self.waveform.model_dump_json(indent=2).encode())
+        b.seek(0)
+        return b
+
     async def insert_waveform(self) -> None:
```
Will-Cross1, line 34: Is this still supposed to be an async function?

operationsgateway_api/src/records/waveform.py
```diff
@@ -79,19 +84,38 @@ def _create_plot(self, buffer) -> None:
...
+    def get_full_path(relative_path: str) -> str:
+        """
+        Converts a relative waveform path to a full path by adding the 'prefix' onto a
+        relative path of a waveform. The full path doesn't include the bucket name
+        """
+        return f"{Waveform.echo_prefix}/{relative_path}"
 
-        if waveform_data:
+    @staticmethod
+    async def get_waveform(waveform_path: str) -> WaveformModel:
```
Will-Cross1, line 103: Does get_waveform need to be asynchronus?

operationsgateway_api/src/records/waveform.py
Will-Cross1, line 34: Is this still meant to be an asynchronus function?

operationsgateway_api/config.yml.example
Will-Cross1, line 20: upload_image_threads should still exist in the example config

Will-Cross1 (changes requested):
Looks good apart from these three minor things

operationsgateway_api/src/records/waveform.py
MRichards99: Fixed in 861ebc2e4d86e1d6bd48ef1eba48cb77c4a98c7d

operationsgateway_api/src/records/waveform.py
MRichards99: Fixed in 861ebc2e4d86e1d6bd48ef1eba48cb77c4a98c7d

operationsgateway_api/config.yml.example
MRichards99: Fixed in 861ebc2e4d86e1d6bd48ef1eba48cb77c4a98c7d

MRichards99:
@Will-Cross1 Thanks for reviewing, I've made the suggested changes and marked the PR for re-review. Could you take another look please?

Will-Cross1 (approved):
Looks good

PR #101, opened 2024-03-04
DSEGOG-301 Add documentation

MRichards99:
This PR adds some documentation to `docs/` as it's been noted that there are some areas of knowledge that would benefit if they were written down.

There's only one small code change with the old ingestion script (dropping the channels collection to ensure moving between test data sets works) so have a read through the documentation, check if it makes sense and whether there's anything that needs adding that I've missed.


Will-Cross1 (approved):
Looks good

PR #102, opened 2024-03-04
DSEGOG-304 Add flag to allow an API instance to be pointed to https

MRichards99:
This is a small PR to pre-empt an issue that I might find when re-ingesting the data on the dev server. It allows you to work with instances of the API which use https, something which the dev server now does.

This change has only been made on the new script. It doesn't need to be made on `ingest_hdf.py` as when you don't want the script to launch an instance of the API (i.e. you want to use an instance already alive), you specify the entire URL using the `-u` option (which includes the http part)


Will-Cross1 (approved):
Looks good

PR #103, opened 2024-03-04
Fix scaling by factor of 257 -> 256 in apply false colour

patrick-austin:
~Pending further changes~
Autoscaling or other changes to be done at a later date.


kevinphippsstfc (approved):
Looks good. Thanks for taking the time to correct this and explaining the finer details to me!

PR #106, opened 2024-03-22
DSEGOG-262 echo hdf tests

Will-Cross1:
Added pytests to the code that changed in DSEGOG-267 HDF validity checks branch to increase the overall code coverage


test/records/test_echo_interface.py
```diff
@@ -103,3 +143,80 @@ def test_upload_file_object(self, _):
...
+
+            mock_bucket.upload_fileobj.side_effect = ClientError(
+                {"Error": {"Code": "AccessDenied", "Message": "Access Denied"}},
+                "upload_fileobj",
+            )
+
+            test_echo = EchoInterface()
+
+            with pytest.raises(EchoS3Error, match="when uploading file at"):
+                test_echo.upload_file_object(BytesIO(), self.test_image_path)
```
MRichards99, line 158: Similar to above, as this is testing for things going wrong, this should be in a separate test function. Rename this one to test_valid_upload_file_object() and create a new function called test_invalid_upload_file_object().

test/records/test_echo_interface.py
```diff
@@ -103,3 +143,80 @@ def test_upload_file_object(self, _):
...
+                "delete",
+            )
+
+            test_echo = EchoInterface()
+            with patch(
+                "operationsgateway_api.src.records.echo_interface.log.info",
+            ) as mock_log_info:
+                test_echo.delete_file_object(self.test_image_path)
+
+                assert mock_log_info.call_count == 1
```
MRichards99, line 222: Similar to above, as this is testing for things going wrong, this should be in a separate test function.

MRichards99 (changes requested):
The tests you've added look good, they just need splitting into separate test functions so there's a distinction between valid and invalid test cases :)

test/records/test_echo_interface.py
```diff
@@ -45,6 +49,29 @@ def test_valid_init(self, mock_resource):
...
+        with patch("boto3.resource") as mock_resource:
+            mock_bucket_instance = mock_resource.return_value.Bucket.return_value
+
+            mock_bucket_instance.creation_date = None
+
+            with pytest.raises(
+                EchoS3Error,
+                match="Bucket for image storage cannot be found",
+            ):
+                test_echo = EchoInterface()
```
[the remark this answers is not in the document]
Will-Cross1: resolved in push

test/records/test_echo_interface.py
Will-Cross1: resolved in next commit

test/records/test_echo_interface.py
Will-Cross1: resolved in next commit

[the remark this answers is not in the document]
Will-Cross1: I set it to check if there was one log call because there was one at the top of the function


its testing that in this piece of code
```
except ClientError as e:
    if e.response["Error"]["Code"] == "404":
        log.info("%s Deletion successful.", image_path)
else:
    log.error(
```

nothing is ran (the response code wan't 404)

I am not sure how I would test it differently

MRichards99 (approved):

PR #110, opened 2024-04-08
Use plt.figure in place of plt.rcParams for figsize

patrick-austin:
Test failure will be resolved by #109


MRichards99 (approved):
Looks good, makes the code a bit easier to read - I wrote the code originally and I think I lacked knowledge that you could do this. I must've found an example that used `rcParams` and just went to use that!

PR #112, opened 2024-04-09
DSEGOG-299 Implement Favourite Filters

MRichards99:
This PR adds the favourite filters functionality, which allows users to have CRUD functionality for favourite filters. From a backend perspective, favourite filters are simply named strings that belong to specific user in the `users` collection. In the frontend, these will act as a way for data to be filtered (e.g. only display records where data channel X > 5). In the backend, we don't need to know this in terms of functionality, we're just storing and fetching these things.

There are 5 endpoints added in this PR:
- `POST /users/filters` - save a favourite filter for the current user
- `PATCH /users/filers/{id_}` - update a specific user's favourite filter (update name/the filter itself)
- `GET /users/filters` - get all of a user's favourite filters and return them as a list
- `GET /users/filters/{id_}` - get a specific favourite filter
- `DELETE /users/filters/{id_}` - delete a favourite filter from the user's document in the database

For each endpoint, the current user is worked out based on the contents of the bearer token.

The functionality is in the `FavouriteFilter` class - this is a relatively straightforward CRUD class that uses `MongoDBInterface` to contact the database. As the data is being stored in an array (called `filters`) in a user document, the MongoDB `$push` and `$pull` operators are being used for the first time in the codebase. These are being used to add & remove a filter from the array respectively.

When a filter is deleted, `update_one()` must be called as you're effectively updating the user collection to not have the filter any more. When I wrote the delete and update functionality in `FavouriteFilter`, I found the error handling was very similar across both methods so I chose to extract it out into `_process_update_result()`. I created a simple enum `FilterCRUDOperation` so when errors are logged, it's clear to see whether this came from an update or a delete.

When retrieving data, methods return using the `FavouriteFilterModel`, a new Pydantic model.

In terms of tests, there's a mixture of unit tests (for `FavouriteFilter`) which database operations are mocked and some integration tests on the endpoints themselves where database operations are done for real and are setup/cleaned up by pytest fixtures.


MRichards99:
The CI is failing due to a single test failure that will be fixed once #109 has been merged

MRichards99:
Just merged main in so the CI passes (safety job was failing)

operationsgateway_api/src/routes/filters.py
```diff
@@ -0,0 +1,176 @@
...
+    "/users/filters",
+    summary="Store a favourite filter into the user's document in the database",
+    response_description="List of favourite filters",
+    tags=["Filters"],
+)
+@endpoint_error_handling
+async def create_favourite_filter(
+    access_token: AuthoriseToken,
+    name: str = Query(
+        "",
```
patrick-austin, line 76: Should `name` be optional with a default of `""`? Is it meaningful to create a favourite filter without a name (or should it always be required)?

operationsgateway_api/src/routes/filters.py
```diff
@@ -0,0 +1,176 @@
...
+        description="Name of the favourite filter",
+    ),
+    filter: str = Query(  # noqa: A002
+        "",
```
patrick-austin, line 80: Should `filter` be optional with a default of `""`? Is it meaningful to create a favourite filter without a filter (or should it always be required)?

operationsgateway_api/src/routes/filters.py
```diff
@@ -0,0 +1,176 @@
...
+    access_token: AuthoriseToken,
+    _id: Annotated[
+        str,
+        Path(
+            ...,
+            description="`_id` of the favourite filter to update",
+        ),
+    ],
+    name: str = Query(
+        "",
```
patrick-austin, line 120: Again I'm not sure of the implications of this defaulting to `""`. How can a user change only the `name`, for example? If they omit the `value` it would default to `""` and so overwrite the value they wanted to persist with an empty string?

operationsgateway_api/src/routes/filters.py
```diff
@@ -0,0 +1,176 @@
...
+        description="Name of the favourite filter",
+    ),
+    filter: str = Query(  # noqa: A002
+        "",
```
patrick-austin, line 124: Again I'm not sure of the implications of this defaulting to `""`. How can a user change only the `name`, for example? If they omit the `value` it would default to `""` and so overwrite the value they wanted to persist with an empty string?

test/endpoints/favourite_filters/conftest.py
```diff
@@ -0,0 +1,108 @@
...
+
+    yield inserted_filter
+
+    test_filter.delete(inserted_filter["_id"])
+
+
+@pytest.fixture()
+def multiple_favourite_filters():
+    test_filter = FavouriteFilter()
+    test_filter.delete_all()
```
patrick-austin, line 21: Why do we need this in the setup of the fixture (before the `yield`)? If another test has generated a favourite filter that conflicts with this fixture then it should be the responsibility of that test to clear itself up (otherwise, if more tests are added and the order changes, or tests are selectively run then we can't guarantee that this `delete_all` will occur in the "right" place).

test/endpoints/favourite_filters/test_save_filter.py
```diff
@@ -0,0 +1,32 @@
...
+    def test_save_filter(
+        self,
+        test_app: TestClient,
+        login_and_get_token,
+    ):
+        name = "Test Create Filter"
+        filter_content = "PM-201-HJ-CRY-FLOW > 10"
+
+        test_response = test_app.post(
+            f"/users/filters?name={name}&filter={filter_content}",
```
patrick-austin, line 17: We're only testing this when `name` and `filter` is set, when (as mentioned on the endpoint) they will also default to `""` if not set, resulting in (e.g.) `[{'_id': ObjectId('6655fb00434b426cbaed6370'), 'filter': '', 'name': ''}]`. If this is indeed intentional, then we should test that this functionality is correct with an additional test case where the query params are not set.

test/endpoints/favourite_filters/test_save_filter.py
```diff
@@ -0,0 +1,32 @@
...
+        assert test_response.status_code == 201
+
+        # Testing the response is a valid object ID
+        filter_id = test_response.json()
+        ObjectId(filter_id)
+
+        test_filter = FavouriteFilter()
+        result = test_filter.find_one(filter_id)
+        print(f"Result 123: {result}")
+        assert result.get("filters") is not None
```
patrick-austin, line 30: In light of the point above about defaulting to empty strings, I think it would be better here if we asserted on the contents, so we can be sure that the behaviour is desirable. E.G. rather than just asserting `is not None`:
```suggestion
    assert len(result.get("filters")) == 1
    assert "_id" in result.get("filters")[0]
    assert result.get("filters")[0]["name"] == "Test Create Filter"
    assert result.get("filters")[0]["filter"] == "PM-201-HJ-CRY-FLOW > 10"
```

test/endpoints/favourite_filters/test_save_filter.py
```diff
@@ -0,0 +1,32 @@
...
+
+        test_filter.delete(filter_id)
```
patrick-austin, line 32: If the test fails on an assertion, this line won't be reached. Either this should be in a fixture, or we should wrap the test in `try` and put the deletion in a `finally` block to ensure it's always run.

test/endpoints/favourite_filters/test_update_filter.py
```diff
@@ -0,0 +1,35 @@
...
+        single_favourite_filter,
+        test_app: TestClient,
+        login_and_get_token,
+    ):
+        updated_name = "Test Update Filter"
+        updated_filter = "PM-201-HJ-CRY-FLOW > 20"
+
+        test_response = test_app.patch(
+            f"/users/filters/{single_favourite_filter['_id']}?name={updated_name}"
+            f"&filter={updated_filter}",
```
patrick-austin, line 19: As mentioned above, we are currently only testing a case when both query params are set, but we should also be testing when only one is set and when neither are set. Especially as this is labelled as a `patch` endpoint, it should **not** require the entire document to be submitted as with post, which we aren't testing.

patrick-austin (changes requested):
Some minor changes like opportunities for type aliasing and copy paste errors in the endpoint metadata.

My main worry is the use of `""` as a default for the `post` and `patch` endpoints.

In the former, it's plausible you may want to create a filter where these both default to `""`, but IMHO it would make more sense for them both to be required.

In the latter case, the fact that either `name` or `filter` being missing will lead to that field being wiped is incompatible with the endpoint being labeled `patch`. Either:
- Both should be required and it should be relabeled as `post`
- We should default to `None` rather than `""` and then only update the fields if `is not None` (i.e. give ourselves a way to distinguish between the query params being set or unset - which isn't possible when they default to `""`)

operationsgateway_api/src/mongo/interface.py
```diff
@@ -102,7 +102,7 @@ async def find_one(
         collection_name: str,
         filter_: Dict[str, Any] = None,
         sort: List[Tuple[str, int]] = None,
-        projection: List[str] = None,
+        projection: Optional[Union[Mapping[str, Any], Iterable[str]]] = None,
```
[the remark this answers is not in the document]
MRichards99: Good point - fixed in 6625e2e09d334593ca0bfd28548df3cdfc946c55

operationsgateway_api/src/mongo/interface.py
```diff
@@ -49,7 +49,7 @@ def find(
         skip: int = 0,
         limit: int = 0,
         sort: Union[str, List[Tuple[str, int]]] = "",
-        projection: Union[None, List[str]] = None,
+        projection: Optional[Union[Mapping[str, Any], Iterable[str]]] = None,
```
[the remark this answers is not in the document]
MRichards99: Good point - fixed in 6625e2e09d334593ca0bfd28548df3cdfc946c55

operationsgateway_api/src/models.py
```diff
@@ -245,3 +245,9 @@ class UserSessionListModel(UserSessionModel):
     # displaying on output
     username: Optional[str] = Field(None, exclude=True)
     session: Optional[Dict[str, Any]] = Field(None, exclude=True)
+
+
+class FavouriteFilterModel(BaseModel):
+    id_: Optional[Annotated[ObjectId, PyObjectId]] = Field(None, alias="_id")
```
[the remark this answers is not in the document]
MRichards99: Good point - fixed in 6625e2e09d334593ca0bfd28548df3cdfc946c55

operationsgateway_api/src/routes/filters.py
MRichards99: Good point, this endpoint is intended to have both query parameters be mandatory. I've fixed this in 1c493c6e455dcbd9a900d7eb689c5381bc6fc603

operationsgateway_api/src/routes/filters.py
MRichards99: Good point, this endpoint is intended to have both query parameters be mandatory. I've fixed this in 1c493c6e455dcbd9a900d7eb689c5381bc6fc603

test/endpoints/favourite_filters/test_save_filter.py
MRichards99: No, this functionality isn't intended - I've added two test cases to test the intended functionality in 1c493c6e455dcbd9a900d7eb689c5381bc6fc603

test/endpoints/favourite_filters/test_save_filter.py
MRichards99: Good spot - a fixture isn't appropriate as you can't pass data to them and you don't know what `filter_id` will be until the test executes. I've used the `finally` syntax, the change is in a03399c144cca1fd96d0954cb5ec276f5b8aca18

operationsgateway_api/src/routes/filters.py
MRichards99: I've refactored this endpoint so you can partially update a filter and when nothing is updated, nothing is done. See 4c3feec77ac8e923e65a991ad1deb888e98caab7

operationsgateway_api/src/routes/filters.py
MRichards99: Similar to my previous comment, I've refactored this endpoint to more sensibly behave when updating filters

test/endpoints/favourite_filters/conftest.py
MRichards99: This is a defensive measure to protect against filters in the database that shouldn't be - you can't guarantee someone hasn't been in the database and fiddled before running the tests! I've looked through the tests and I believe every test does clean up the filters it creates.

test/endpoints/favourite_filters/test_update_filter.py
MRichards99: The refactor now allows partial updates and adds some test cases to test updating just the filter, just the name etc. See 4c3feec77ac8e923e65a991ad1deb888e98caab7

MRichards99:
Pushed some further changes to improve the behaviour - I've requested a re-review. I also spotted a bug when sending a request to `GET /users/filters` where the user doesn't have any filters (it returned 500 instead of an empty list). I've fixed that bug in 9fa695d51a9975ee25250fdbbb6c4d8a89d8f023

operationsgateway_api/src/routes/filters.py
```diff
@@ -130,16 +128,23 @@ async def update_favourite_filter(
...
+        return "No updates necessary"
+
     token_payload = JwtHandler.get_payload(access_token)
     username = token_payload["username"]
 
     log.info("Updating favourite filter '%s' for user '%s", _id, username)
     log.debug("Name: %s, Filter: %s", name, filter)
 
+    filter_result = await FavouriteFilter.get_by_id(username, _id)
+    updated_name = name if name is not None or name == "" else filter_result.name
```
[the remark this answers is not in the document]
patrick-austin: Alternatively, maybe this should be checking truthyness like on line 131. Especially since trying to set name to `""` will fail once it hits:
https://github.com/ral-facilities/operationsgateway-api/blob/b653096244e07c8666240c67a35bdf5eef84e8f4/operationsgateway_api/src/users/favourite_filter.py#L44-L45

Currently, it is possible to set the filter to an empty string by:
```bash
curl -X 'PATCH'   'http://127.0.0.1:8000/users/filters/6655fb00434b426cbaed6370?name=n&filter='   -H 'accept: application/json'   -H 'Authorization: Bearer ...'
```
Since `name` being set allows us to pass the check on 131, and `"" is not None`. But it is not possible by:
```bash
curl -X 'PATCH'   'http://127.0.0.1:8000/users/filters/6655fb00434b426cbaed6370?&filter='   -H 'accept: application/json'   -H 'Authorization: Bearer ...'
```
Since `name is None` and `filter` evaluates to False. Not sure which is the intentional behaviour here.

patrick-austin (changes requested):
One place where I think we can remove some unreachable logic, but then it got me thinking about the checks of the update filter endpoint more generally. It's minor compared to the changes that have already been made but thought I'd point it out as potentially unexpected edge case behaviour.

operationsgateway_api/src/routes/filters.py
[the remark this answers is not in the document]
MRichards99: I've pushed d40acb4b553742c87a8207a2a001adfd94fd4bd7 so filters cannot be set to an empty string. Good spot about `name == ""`, I think I thought made a difference but I must've saved the file at the wrong time as it's obvious now that it wouldn't get evaluated.

patrick-austin (approved):
Tried the same curl commands as before and with this change I wasn't able to get it to set either of the fields to empty strings, so all the checks seem to do what they should.

PR #113, opened 2024-04-11
DSEGOG-307 Kill API Instances Cleanly in Echo Ingest Script

MRichards99:
This is a small change to improve the Echo ingestion script. It used to not kill the API at the end of the script so the processes needed to be manually killed - the problem would often occur when you launch the API using `poetry run` and it would complain the address is already in use.

To make this better, the kill function now uses `lsof` to pinpoint all processes attached to the port used for the API, puts those PIDs into a list and uses the standard library `os.kill()` to kill them one by one.


MRichards99:
Just merged main in so the CI passes (safety job was failing)

patrick-austin (approved):

PR #118, opened 2024-04-29
DSEGOG-315 MongoDB 5.0 to 7.0

MRichards99:
This is a small PR which helps to move MongoDB 5.0 to 7.0. This changes the GitHub Actions to install MongoDB 7.0 and there's also a guide that allows developers to make this change on their development environments. This change is happening because Database Services are now hosting a database using MongoDB 7.0, so we should move to that version as developers for consistency.


patrick-austin (commented):
Will come back to this and test the commands work for me on my dev environment.

It doesn't matter because it's symlinked but if these commands are for a Rocky8 machine it will use `dnf` rather than `yum` right? Not a request to change just trying to remind myself which way round it is more than anything...

docs/mongodb.md
```diff
@@ -0,0 +1,37 @@
...
+yum erase mongodb-*
+# Check that everything's been uninstalled
+yum list installed | grep -i mongod
+# Remove data directories, databases, logs
+rm -rf /var/log/mongodb /var/lib/mongo*
+# Disable yum repo for MongoDB 5.0
+yum-config-manager --disable mongodb-org-5.0
+
+# Add yum repo for MongoDB 7.0
+touch /etc/yum.repos.d/mongodb-org-7.0.repo
```
patrick-austin, line 19:
```suggestion
sudo systemctl stop mongod
# Uninstall packages using yum
sudo yum erase mongodb-*
# Check that everything's been uninstalled
yum list installed | grep -i mongod
# Remove data directories, databases, logs
sudo rm -rf /var/log/mongodb /var/lib/mongo*
# Disable yum repo for MongoDB 5.0
sudo yum-config-manager --disable mongodb-org-5.0

# Add yum repo for MongoDB 7.0
sudo touch /etc/yum.repos.d/mongodb-org-7.0.repo
```

Commands won't run as non-root.

docs/mongodb.md
```diff
@@ -0,0 +1,37 @@
...
+
+# Paste the following into /etc/yum.repos.d/mongodb-org-7.0.repo
+# Edit the /8/ to /9/ if you're using Rocky 9
+[mongodb-org-7.0]
+name=MongoDB Repository
+baseurl=https://repo.mongodb.org/yum/redhat/8/mongodb-org/7.0/x86_64/
+gpgcheck=1
+enabled=1
+gpgkey=https://pgp.mongodb.com/server-7.0.asc
```
patrick-austin, line 28: Since this isn't a bash command I think it would be better to make it a separate code block, e.g (formatting isn't great because of the nested backticks but hopefully it makes sense:
```suggestion

Paste the following into /etc/yum.repos.d/mongodb-org-7.0.repo (edit the /8/ to /9/ if you're using Rocky 9):
```ini
[mongodb-org-7.0]
name=MongoDB Repository
baseurl=https://repo.mongodb.org/yum/redhat/8/mongodb-org/7.0/x86_64/
gpgcheck=1
enabled=1
gpgkey=https://pgp.mongodb.com/server-7.0.asc
```

```bash
```

docs/mongodb.md
```diff
@@ -0,0 +1,37 @@
...
+
+# Install the latest stable version of MongoDB
+yum install mongodb-org
+# Start MongoDB, enter the shell and check the "Using MongoDB:" line lists a version 7.x.x
+systemctl start mongod
```
patrick-austin, line 33:
```suggestion
sudo yum install mongodb-org
# Start MongoDB, enter the shell and check the "Using MongoDB:" line lists a version 7.x.x
sudo systemctl start mongod
```

patrick-austin (changes requested):
Mostly just a bunch of commands that need sudo to run. Main point is that this won't work without the changes from #120 (I assume).

docs/mongodb.md
MRichards99: I've committed all the suggested changes you proposed, thanks for making them. I had to play around with the formatting on this suggestion and settled on the changes in [f82776c](https://github.com/ral-facilities/operationsgateway-api/pull/118/commits/f82776cee9023178576dc8dee6d3fba0568dd322). Can you take a look and make sure you're happy with how it looks please?

docs/mongodb.md
```diff
@@ -0,0 +1,37 @@
...
+mongosh
```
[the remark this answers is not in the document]
MRichards99: Good point, let's leave this unresolved so I remember to not merge this until the other PR is merged.

docs/mongodb.md
```diff
@@ -5,32 +5,34 @@ These instructions allow developers to move from MongoDB 5.0 to 7.0 to match the
...
+```ini
 [mongodb-org-7.0]
 name=MongoDB Repository
 baseurl=https://repo.mongodb.org/yum/redhat/8/mongodb-org/7.0/x86_64/
 gpgcheck=1
 enabled=1
 gpgkey=https://pgp.mongodb.com/server-7.0.asc
-
+```
+```
```
patrick-austin, line 31:
```suggestion
```bash
```

Really minor, we can use the `bash` language support for this block like the first one (practically I think the only difference is it will gray out the comments). Appreciate this was probably missing from my suggested change because it struggles with nesting back ticks inside the `suggestion` backtick box.

patrick-austin (changes requested):
Another incredibly pedantic change request (the real problem is I think the language support / syntax highlighting feature is really cool so I can't overlook and opportunity to use it- even when it _really_ doesn't matter...)

MRichards99:
Just re-requested a re-review - fixed the Safety job that was failing before

patrick-austin (approved):
Reading back over my old comments, the changes were good, just the reference to `mongosh` was dependent on #120 which I've just approved, so approving this as well.

PR #119, opened 2024-04-29
DSEGOG-305 Config & Documentation for Data Re-ingestion

MRichards99:
This is a small PR that makes a couple of minor changes for the re-ingestion of data on the new database that Database Services host. I've made a couple of updates to the documentation (mainly an update to the dev server docs to explain the current situation with the data on the machine). For the ingestion script, I've moved from the configuration using a hostname and port for the database to a connection string. I've made this change because the new connection string from Database Services is quite complex and therefore the string is a better fit than the hostname and port.


.github/ci_ingest_echo_config.yml
```diff
@@ -7,8 +7,7 @@ ssh:
   enabled: false
   ssh_connection_url: 127.0.0.1
 database:
-  hostname: localhost
-  port: 27017
+  connection_uri: mongodb://localhost:27017/opsgateway
   name: opsgateway
```
patrick-austin, line 11: As far as I can see, we're only still using this in one place:
https://github.com/ral-facilities/operationsgateway-api/blob/1733055c5a068d5444c18fb0ac65258539dadd19/util/realistic_data/ingest_echo_data.py#L39
(and two places for print statements, which I'm less concerned about)
https://github.com/ral-facilities/operationsgateway-api/blob/1733055c5a068d5444c18fb0ac65258539dadd19/util/realistic_data/ingest/local_command_runner.py#L46
https://github.com/ral-facilities/operationsgateway-api/blob/1733055c5a068d5444c18fb0ac65258539dadd19/util/realistic_data/ingest/ssh_handler.py#L32

In most places since we're using `mongoimport` as a subprocess I gather the fact that `connection_uri` ends `/opsgateway` means we don't need to pass it (the DB name) directly. So I'm sort of wondering if we need to still have `Config.config.database.name` at all or if for that one case we still need it we could split it from the last `/` in the uri (as apparently `pymongo.errors.InvalidName: database names cannot contain the character '/'`)?

Alternatively, given that in the main config we have:
https://github.com/ral-facilities/operationsgateway-api/blob/1733055c5a068d5444c18fb0ac65258539dadd19/operationsgateway_api/config.yml.example#L23-L25
We could make this more consistent and keep the `name` and don't have it trailing the port.

The motivation of this change was around the connection string becoming to complex, so maybe the current approach is the most appropriate (without knowing the real string it's hard to judge). In reality this is all minor, but I suppose my concern would be that you change the `name` but not the `connection_uri` and now have one command that tries to use a different DB to the other (or vice versa). Ideally I think we'd have a single source of truth for `opsgateway`, regardless of whether it's via `connection_uri` or `name` - but ideally not both?

MRichards99:
Requested a re-review as I feel I've addressed the PR comment by the changes made in #129

patrick-austin (approved):
Yeah I think #129 covers all the points here

PR #123, opened 2024-05-14
DSEGOG-317 Bug fix for updating records

MRichards99:
This PR fixes a bug I found when records are updated but the images/waveforms aren't present on Echo. The channels were skipped because the database says the channel exists but this doesn't always mean the data itself exists on Echo. An additional check has been added for image and waveform channels to see if it's actually on Echo or not. If the file isn't present on Echo, the channel is ingested (meaning the file will end up on Echo).


MRichards99:
Safety job on CI is failing which will be fixed when #122 is merged

operationsgateway_api/src/records/ingestion/partial_import_checks.py
```diff
@@ -92,15 +100,41 @@ def channel_checks(self):
...
+                        path = Waveform.get_full_path(channel_model.waveform_path)
+
+                    object_stored = self._is_image_or_waveform_stored(path)
+
+                if object_stored:
+                    rejected_channels[channel_name] = (
+                        "Channel is already present in existing record"
+                    )
+                else:
+                    accepted_channels.append(channel_name)
```
patrick-austin, line 118: I don't think this will work for non-Image/Waveform channels. `object_stored` is only set on line 111, which can only be reached if the channel model `isinstance(channel_model, (ImageChannelModel, WaveformChannelModel))`. So if that's `False` we're either going to come into this without `object_stored` being set leading to a `NameError` or we use the value of the previous iteration (i.e. the last Image/Waveform we encountered).

Haven't tested this (syntax/functionality might be broken), but I think we'd want something like:

```suggestion
            if channel_name in stored_channels:
                if isinstance(channel_model, ImageChannelModel):
                    path = Image.get_full_path(channel_model.image_path)
                    object_stored = self._is_image_or_waveform_stored(path)
                elif isinstance(channel_model, WaveformChannelModel):
                    path = Waveform.get_full_path(channel_model.waveform_path)
                    object_stored = self._is_image_or_waveform_stored(path)
                else:
                    object_stored = True

                if object_stored:
                    rejected_channels[channel_name] = (
                        "Channel is already present in existing record"
                    )
                else:
                    accepted_channels.append(channel_name)
```

patrick-austin (changes requested):
I don't think this will work for non-Image/Waveform channels (but have only looked at this in the GH web UI, not tried to actually run it so might be wrong...)

operationsgateway_api/src/records/ingestion/partial_import_checks.py
MRichards99: Good spot, I've managed to break this when updating a HDF file that contains no image or waveforms channels (an edge case I didn't consider). `UnboundLocalError` was the precise exception I encountered. I'll apply the fix tomorrow.

MRichards99: I've just pushed a fix for this - it's pretty much your suggestion but I had to add an additional else statement at the end to fix a bug where accepted channels were no longer appearing in the response.

patrick-austin (approved):
Change looks good to me, glad to hear it works with a image/waveform-less import.

PR #129, opened 2024-06-21
DSEGOG-326 Ingestion script refactoring

MRichards99:
This PR addresses some points picked up from #120 and #119 to improve the Echo ingestion script:
- I've removed the `database.name` config so the database configuration is only in `database.connection_uri`. This means there is now a single source of truth in terms of the database name, which is referenced in the URI itself. Comment from PR talking about this: https://github.com/ral-facilities/operationsgateway-api/pull/119#discussion_r1611596576
- Removed `SSHHandler` because that functionality simply isn't used. I refactored `LocalCommandHandler`, moving most of the functionality into a new class, `DatabaseOperations`. The  comment from PR talking about this: https://github.com/ral-facilities/operationsgateway-api/pull/119#discussion_r1619100228
- Stopped using `mongoimport` via `subprocess` and using `pymongo` to communicate with the database - this has removed the need for `sleep(5)` when dropping database collections (see https://github.com/ral-facilities/operationsgateway-api/pull/120#discussion_r1617168428)


util/realistic_data/ingest/s3_interface.py
```diff
@@ -34,15 +35,33 @@ def download_manifest_file(self) -> BytesIO:
...
+            )
+            print(
+                f"{Config.config.database.remote_experiments_file_path} hasn't been"
+                " created yet..",
+            )
+            sleep(1)
+
+        print(
+            f"Written '{Config.config.database.remote_experiments_file_path}' to file",
+        )
```
patrick-austin, line 64: Don't have a problem with this, since we were doing it before in `LocalCommandRunner` so I'm sure there's a good reason, but I am curious as to what that reason is. How can the file not exist if we open, write and close it before reaching this loop? Is this just a general Python thing (allowing the the next lines to run before the file has actually been "saved")?

patrick-austin (approved):
The changes made look good, I just can't work out the reason behind the while loop (so I'm happy to approve it but would love to know why we need it).

PR #132, opened 2024-07-15
DSEGOG-334 Fix bug relating to experiments on ingest script

MRichards99:
I've fixed the bug that imports the experiments but immediately wipes the database afterwards - I've just swapped the order of these two things. I've also removed the skip on the test that wouldn't run on GitHub Actions as it is no longer relevant - the test doesn't rely on the dev scheduler so can be run on CI. I removed a couple of commented out skips too.

To test, I ran the ingest script with the `wipe_database` config option set to true and then ran the tests.

### DIFF
--- a/test/endpoints/test_get_experiments.py
+++ b/test/endpoints/test_get_experiments.py
@@ -1,20 +1,8 @@
-import os
-
 from fastapi.testclient import TestClient
-import pytest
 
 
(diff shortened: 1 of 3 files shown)


patrick-austin (approved):
Re-ingesting with with the changed script fixed the test failure I was getting locally

PR #134, opened 2024-07-17
DSEGOG-337 Fix daily ingestor script trying to ingest the same file twice

MRichards99:
This PR fixes an issue I've noticed on the daily ingestion of simulated data where the ingestion can't quite keep up with the cron job scheduled for every 60 seconds, thereby causing some errors when both try to ingest the same file. When there are multiple files to be ingested by `daily_ingestor.py`, the script loops through each of the files, ingesting them one at a time. Sometimes, ingesting all of the files in the list takes over 60 seconds. As such, the second execution of the script can try to ingest files that the first execution is ingesting which can throw up errors because the file can’t be found (as the first execution has finished the ingestion by that point and deletes the file).

This is resolved by creating a temporary directory where the files due to be ingested are stored so no other executions of the script can access them - kind of like a lock in a way? This directory is cleaned up and removed at the end of the script and the existing behaviour with unsuccessful ingestions hasn't changed.

I've added a few comments in the script so hopefully it makes sense. To test this, I have deployed it on the dev server and it has acted in the way that I hoped it would.

### DIFF
--- a/util/realistic_data/daily_ingestor.py
+++ b/util/realistic_data/daily_ingestor.py
@@ -55,8 +55,12 @@ def main():
                 f" Response: {response_code}. Moving this file to so it can be"
                 f" investigated by a human: {failed_ingests_directory}",
             )
-
-            file_to_ingest.rename(f"{failed_ingests_directory}/{file_to_ingest.name}")
+            try:
+                file_to_ingest.rename(
+                    f"{failed_ingests_directory}/{file_to_ingest.name}",
+                )
+            except FileNotFoundError as e:
+                print(e)
 
 
 if __name__ == "__main__":
--- a/util/realistic_data/daily_ingestor.py
+++ b/util/realistic_data/daily_ingestor.py
@@ -28,11 +28,25 @@ def main():
             tzinfo=tz.gettz("Europe/London"),
         )
         for file in hdf_files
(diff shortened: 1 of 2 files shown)


util/realistic_data/daily_ingestor.py
```diff
@@ -28,35 +28,83 @@ def main():
...
+
+    in_progress_date_format = f"{hdf_file_date_format}%S"
+    current_datetime = (
+        datetime.now().replace(microsecond=0).strftime(in_progress_date_format)
+    )
+    Path(f"{hdf_data_directory}/in_progress_{current_datetime}").mkdir(
+        parents=True,
+        exist_ok=True,
+    )
+    in_progress_dir = Path(f"{hdf_data_directory}/in_progress_{current_datetime}")
```
kevinphippsstfc, line 43: Could this go above line 39? 
Then you can do `in_progress_dir.mkdir` on line 39 without repeating the concatenation of the file path.

kevinphippsstfc (changes requested):

util/realistic_data/daily_ingestor.py
MRichards99: Good spot! That means `mkdir()` can go onto one line so that's nice :)

kevinphippsstfc (approved):

PR #138, opened 2024-07-22
Fix performance of images (and export)

kevinphippsstfc:
Originally we had assumed that the performance issue could be due to the move from local file storage to using Echo, but having added some timing debug statements I found that my test 3.4MB image (20230605080000/CM-202-CVC-CAM-1) was actually being returned very quickly from Echo in about 0.1 seconds. 

Analysing the breakdown of the request timings in Postman showed that the returning of the image over HTTP was the slow bit and with a bit of googling I found this on StackOverflow:

https://stackoverflow.com/questions/55873174/how-do-i-return-an-image-in-fastapi

which warns against using StreamingReponse because of the way it chunks up the response. (I can't remember why I used StreamingResponse in the first place - probably it was used in an example I based the code on).

So changing to use the basic fastapi Response seemed to speed things up.

The following request in Postman:
{{og-api}}/images/20230605100000/CM-202-CVC-CAM-1?upper_level=5&colourmap_name=jet_r
which previously took 3.5 to 4 secs was reduced to about 1.5 secs (with half of that being the time to apply false colour)

I also found that I had used StreamingResponse for the Data Export functionality so made the same change there and:
{{og-api}}/export?conditions={"_id": {"$in": ["20230605080000","20230605090000","20230605100000","20230605110000","20230605120000"]}}&projection=channels.FE-204-PSO-CAM-1
which previously took about 11 secs was reduced to about 2 secs, so a big speedup there.


MRichards99 (approved):
I've tested this on my machine and can also see similar performance improvements. The image endpoint spends most of its time converting the image to false colour and given this is a matplotlib function, there's not much else we can do to speed this up. Good performance improvements, should be noticeable on the frontend

PR #139, opened 2024-07-23
DSEGOG-309 Refactor ChannelManifest to use ChannelManifestModel

MRichards99:
This PR refactors `ChannelManifest` so it returns `ChannelManifestModel` from `get_most_recent_manifest()`, instead of a dictionary. It makes use of the existing method `_use_model()` to turn the data from the database into the Pydantic model.

I did the refactor in a couple of steps so it was easier in my head - I created a new function called `get_most_recent_manifest_new()` containing the refactored code and slowly changed all the uses of `get_most_recent_manifest()` to the new function, making changes to additional code as needed (i.e. interacting with the data via the Pydantic model, not a dictionary). Once everything had been converted, I replaced `get_most_recent_manifest_new()` with `get_most_recent_manifest()` and deleted the old code.

### DIFF
--- a/operationsgateway_api/src/channels/channel_manifest.py
+++ b/operationsgateway_api/src/channels/channel_manifest.py
@@ -45,7 +45,7 @@ class ChannelManifest:
         Validate the user's incoming manifest file by comparing that with the latest
         version stored in the database
         """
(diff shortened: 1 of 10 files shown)


MRichards99:
Safety failure fixed in #138

kevinphippsstfc (approved):

PR #140, opened 2024-08-14
DSEGOG-321 Fully implement delete records

MRichards99:
This PR adds full implementation to the `DELETE /records/{_id}` endpoint by deleting images and waveforms stored on Echo. Previously, the record would be removed from the database but images & waveforms associated with that record would still be stored in Echo.

To do this, a new function has been added to the `EchoInterface` class, `delete_directory()`. This allows all of the images or all of the waveforms for a particular record to be removed in one call (two calls total, one for images, one for waveforms). This function is called in the entrypoint function for the `DELETE /records/{_id}` endpoint. I've added tests for this new code.

I fixed a bug occurring with some of the tests in 95321bacc835cbf73e6b76b95fc030f662d91cd2 as I noticed images and waveforms weren't being deleted for a particular record ID. In e7ca92f64b7cf56c44f7f990ee964b2524be4a3d, I improved the integration test for `DELETE /records/{_id}` endpoint by moving the setup and teardown into the fixture (it was in the test itself before).


moonraker595:
Hi, @MRichards99  I'm trying to get this running locally with some test data in the local db/echo instances to be able to test this functionality.
 When running `
 python3 -m util.realistic_data.ingest_echo_data` 
 or 
 `poetry run python util/realistic_data/ingest_echo_data.py` 
 
 I get issues relating to libc.so.6:
`ERROR:   ("dlopen(libc.so.6, 0x0006): tried: 'libc.so.6' (no such file), '/System/Volumes/Preboot/Cryptexes/OSlibc.so.6' (no such file), '/opt/homebrew/lib/libc.so.6' (no such file), '/System/Volumes/Preboot/Cryptexes/OS/opt/homebrew/lib/libc.so.6' (no such file), '/opt/homebrew/lib/libc.so.6' (no such file), '/System/Volumes/Preboot/Cryptexes/OS/opt/homebrew/lib/libc.so.6' (no such file), '/usr/lib/libc.so.6' (no such file, not in dyld cache), 'libc.so.6' (no such file)",)
`

The Readme mentions installing dev tools on linux machines but should I be able to get the `ingest_echo_data` working on my local machine?

MRichards99:
Only Will Cross (an apprentice who worked on OperationsGateway during his rotation to DSEG) has used OperationsGateway API with a Mac - he had various issues with adapting the API to work on his Mac (got it to work in the end), but that's all the interaction this repo has seen with Macs thus far. As a starting point, I'd take a look at this (there's two of the same line in that file):

https://github.com/ral-facilities/operationsgateway-api/blob/fbd8712365c8e0378acfad3900d2376c4df8d179/operationsgateway_api/src/routes/ingest_data.py#L143

It was added to fix a memory leak, but seems to have some relation to the error you're seeing. If those lines are removed, you should test that the memory leak (originally caused by ingestion of the large files) doesn't reappear. Deploying the change onto the dev server and keeping an eye on memory usage according to `systemctl status og-api` (and `htop`) a couple times a day for a week to see whether you the usage is excessive. Remember to disable the cron job that restarts the API at midnight otherwise that will skew your results

moonraker595:
Hi Matt, Thanks for this. I gave up trying to get it to work on my machine and instead got a rocky 8 VM running which I'm using for remote development over ssh, it took a while to set up but seems to be working fine now. I think I'll stick with this going forward.

moonraker595:
After getting everything running, I can see that deleting a record removes its entry in the DB along with the relevant images and waveforms folder in s3. The API then responds correctly with a `204`.

@kevinphippsstfc, what are the next steps here? You're down as the reviewer. Once merged, do we update the dev server, immediately or do we schedule releases?

moonraker595 (approved):
Review in comment section

PR #141, opened 2024-08-14
DSEGOG-311 Move User Office API from SOAP to REST

MRichards99:
This PR moves from the User Office's SOAP API to the REST API. Luckily for us, we only use one endpoint from this API, `/sessions/` which allows us to login. I have removed the creation of a user office SOAP client and replaced the contents of `SchedulerInterface.login()` with code to contact the REST API (using `requests`). There's some error handling in that function that checks the status code of the response and whether the session ID exists in the response.

Similar to the usage of the SOAP client, I have added a decorator called `rest_error_handling` which deals with generic error handling when using requests to access the REST API. These errors are for things like connection errors and timeouts, so a 500 is raised in each case.

Tests have been written for the changes, with 100% coverage for `scheduler_interface.py` and `rest_error_handling.py`! Sam would be proud. I've included the unhappy paths when testing the login and removed mocking of the user office client as we no longer use that.

I've also updated the example config to use the new REST API URL.


moonraker595 (approved):
I've tested this by successfully hitting the `/experiments` endpoint in the API with the updated URL

PR #154, opened 2025-02-07
Update functions help text and add mkdocs for gh-pages

patrick-austin:
- Moved existing developer facing `.md` files to `docs/developer` to avoid ambiguity with user facing documentation
- Added `mkdocs` for building some minimal documentation of the functions which can be accessed outside of the OG frontend
  - In principle this could also include explicit documentation (in addition to the stuff built from docstrings) and more topics, but for the moment this is just focused on functions since that was the scope of 378
- Updated the existing documentation (docstrings and dict that is sent to the frontend) based on feedback from Stephen:
  - Make distinctions between "pixels", "positions" and "values" for functions which return one of these things
  - Functions that deal with peaks will find the global maximum (as opposed to the first or last local maximum)
  - What happens when the "edge" of a peak cannot be found
- Added a `gh-pages` branch where this documentation is hosted: https://ral-facilities.github.io/operationsgateway-api/
- Added new workflow for pushing to the `gh-pages` branch on push to `main`, to automate the building of the documentation. Will have to double check this runs as intended when this branch is merged to main and fix if needed. It also possible to push manually in case we can't get this working, and this is how I made the initial commit to that branch.

Note:
- No changes have been made to the actual implementation of any function
- All descriptions may be modified in a later round of acceptance testing. Any obvious inaccuracies or copy paste errors should be fixed, but I don't think we should spend too much effort on fine tuning the wording as we may need to change it again off the back of acceptance testing anyway.


patrick-austin:
- Changed the description on `index.md`
- Renamed `docs/docs` to `docs/user`. It isn't possible for the `mkdocs.yml` file to be included in the `docs_dir`, which sadly prevents having it as a "sibling" of `index.md`
- Here Functions refers to the user facing feature of being able to apply generic arithmetic and built-in statistical operations to the data stored in OperationsGateway. It is intended to describe the implementation and usage of these Functions to the users. In practice, I am re-using the docstrings from python functions AKA methods, however this is to ensure a single source of truth that can be re-used rather than intending to be a complete API level documentation. For example, there is no documentation here of how the parser works, because the user doesn't really need to know that, just how the selected function will be applied to their data.

PR #155, opened 2025-02-10
Add /maintenance and /maintenance/scheduled GET, POST endpoints

patrick-austin:
Adds two endpoints for getting and setting maintenance and scheduled maintenance messages. This follows this existing format for Sci/DataGateway. These are stored as JSON files rather than in the DB in order to allow the messages to display when the DB is not contactable.
- Added config options and example files for the (scheduled) maintenance
- Added Pydantic models for maintenance and scheduled maintenance bodies
  - Since the issue only listed "info" and "warning" as severity levels expected by the frontend, these are enforced with a `StrEnum`. Can add more cases or relax this to `str` if needed.
- Added POST/GET endpoints for maintenance and scheduled maintenance
  - To cut I/O, these use the [lru_cache](https://fastapi.tiangolo.com/advanced/settings/#lru_cache-technical-details) to cache each file's contents independently. In practice, this means our first GET will read then cache the contents of the file at that time. Subsequent GET requests will use the value cached in memory. The next POST request will clear the cache, such that the next GET will read the updated contents from disk before caching again etc.
- Added tests for new endpoints
  - In order to test the cache (clearing), these run POST and GET requests back to back in the same test. 
- Updated instances of `authorised_route_list`/`authorised_routes` to support these new endpoints.
- Added requests to new collection in Postman

### DIFF
--- a/.github/ci_config.yml
+++ b/.github/ci_config.yml
@@ -4,6 +4,8 @@ app:
   # API will auto-reload when changes on code files are detected
   reload: true
   url_prefix: ""
+  maintenance_file: operationsgateway_api/maintenance.json.example
+  scheduled_maintenance_file: operationsgateway_api/scheduled_maintenance.json.example
 images:
   thumbnail_size: [50, 50]
   default_colour_map: viridis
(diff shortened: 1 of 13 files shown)


moonraker595 (approved):
Looks good & pretty straightforward. I've run the nox sessions which all pass and double-checked the swagger UI. If this ever moves over to docker, we'll have to think about how we store maintenance messages as the container can't store state, but I believe FDP has the same issue so we can see what develops there. Nice work 👍

PR #158, opened 2025-02-12
Use PartialRecordModel instead of dict for Record queries, export, functions

patrick-austin:
- Rename `waveform_variable.py` to `variable_models.py` and add two Pydantic models for internally representing function outputs. These include a `variable_value` field used as a raw input to function evaluations (i.e. `np.ndarray) and a `data` field which is used to store bytes to be returned if we are getting a full Image, Waveform (rather than Records). Neither of these can be included in the usual Records response model, and were previously being added and purged from the `dict` when needed.
  - The `data` stored for images is now `bytes` rather than `BytesIO`, as a result in a few places the use of `getvalue` to access the `bytes` from a `BytesIO` object has been altered, but hopefully these types are now more explicit/intuitive.
- Made `channel_dtyp` a `StrEnum` rather than a `str`. This was part of an initial attempt to use `Discriminators` which ended up being a dead end, but in principle I think it's better to control what values this can take and it should have a minimal, non-functional impact on the code (e.g. `str` equality will act the same as before)
- Added "Partial" representations of Channel and Record models. These are pretty much the same as the full versions, but have two consequences of being populated from Mongo, where projections may have been used:
  - All fields are optional (default to `None`) since the field may not have been in the projections
  - `Any` is not supported, since the values from Mongo should be of the right type (or `None`)
- Replace instances of `dict` representing a Record with the new `PartialRecordModel`
  - As a consequence, lots of `KeyErrors` and accessing via keys replaces with `AttributeErrors` and direct attribute access/use of `getattr`
  - For functions, the `record_id` and an internal representation of the `channels` are used instead of the old `dict`.
  - Return typehints set for `get_records`, `get_record_by_id`. This will now enforce our model so that no unexpected fields are returned, but any field expect the record id can be absent without issue.
- Explicitly test waveform and image functions at the `get_records` endpoint to ensure we won't return `variable_value` or fail Pydantic validation.

EDIT: Codecov isn't loading properly for me (seems to be stuck in some infinite re-direct loop), so I'm struggling a bit to check the coverage reports. Based on running the tests locally, I think most/all the missing branches/lines were already missing back when they were doing things like `record["metadata"] is not None`, but since I've touched those lines to say `record.metadata is not None` it's coming up in the patch. There should be no functional changes here, but regardless I've tried to refactor and improve coverage where I could to be on the safe side.


moonraker595 (commented):
Hi, looks good overall. I just had a minor comment about checking for AttributeErrors. 

Since dictionaries return `None` when using `.get()`, but direct attribute access does not, could there be cases where missing fields raise an uncaught AttributeError?

It looks like `_ensure_waveform_metadata()` handles this for metadata fields, but `channel_name` may still be an issue. For example, in `ExportHandler._process_data_channel`, the current implementation directly accesses `channels[channel_name]`:
```python
channel = channels[channel_name]
```
If channel_name does not exist in channels, I presume this will raise some sort of error? 

Should we be setting and checking it?
```python
channel = channels.get(channel_name, None)
if channel is None:
    # Handle the missing channel
```
I couldn’t find where channel_name is set or if it's validated elsewhere before it reaches this file, so it may not be an issue.

moonraker595:
I can see the coverage report here:
https://app.codecov.io/github/ral-facilities/operationsgateway-api/commits/DSEGOG-232-Use-RecordModel-instead-of-dictionaries
It would be handy if we could run it locally.

patrick-austin:
> Since dictionaries return `None` when using `.get()`, but direct attribute access does not, could there be cases where missing fields raise an uncaught AttributeError?

There are some places (e.g. https://github.com/ral-facilities/operationsgateway-api/pull/158/files#diff-d06e572fbe10b4f01ea1a93081df89c9c9f921aeb4bac5f80f4497f17bb7d0faL282-R293 in `record.py` where existing `KeyError`s have been replaced with `AttributeError`s, but this should only occur when we've got the channel type wrong (in that case, `value` is any `PartialChannelModel` so we could get an error.

Generally, the fact we're using models where every attribute can be, and defaults to, `None` should give the same behaviour as before (if something is missing, accessing it will return `None`). 

> It looks like `_ensure_waveform_metadata()` handles this for metadata fields, but `channel_name` may still be an issue. For example, in `ExportHandler._process_data_channel`, the current implementation directly accesses `channels[channel_name]`:
> 
> ```python
> channel = channels[channel_name]
> ```
> 
> If channel_name does not exist in channels, I presume this will raise some sort of error?
> 
> Should we be setting and checking it?
> 
> ```python
> channel = channels.get(channel_name, None)
> if channel is None:
>     # Handle the missing channel
> ```

If we're talking about https://github.com/ral-facilities/operationsgateway-api/pull/158/files#diff-23281c401ce73b5acfeb05293b5dd087a808ac8594ceee8207316b5e7fb9c946R241 it will still raise a `KeyError` in exactly the same way it did before. [PartialChannels](https://github.com/ral-facilities/operationsgateway-api/pull/158/files#diff-c94a0cf24845812cf802b48249c3461c572f1a00ff74e3e5988764b522f3ff43R169-R172) is just a [type alias](https://typing.readthedocs.io/en/latest/spec/aliases.html) for a `dict`, while most places now just Models because we don't know the channel names we can't use a model for that, and channels are still a `dict` (albeit a `dict[str, PartialChannelModel]` rather than `dict[str, dict]` which they were before.

There may well be a place I've missed where an additional `except AttributeError` is needed in place of `KeyError`, so if there were any other locations you're not sure about please do flag them.

> I can see the coverage report here:

Yeah it's working for me again now (and has been for the past week or so). Think it was a one off/intermittent problem?

> It would be handy if we could run it locally.

I normally just use `--cov --cov-report=term-missing`, but it's a bit of a pain to compare that against the diff to see which missing lines are uncovered. I used to use a VSCode extension which read a local coverage report and highlighted missing lines, I can't remember why I stopped using that. And apparently code cov does support it: https://docs.codecov.com/docs/codecov-uploader (though I haven't read it fully, and would be suspicious of their pricing model). Might be worth looking into if the site goes down again...

patrick-austin:
Have added explicit errors for the cases where a channel name is missing (`channels` without a second term after a `.`) or not recognised, and tests for these errors. Both of these cases would have occurred independently of this change to the record model, as the original `IndexError` and `KeyError` leading to these comes from the representation of channels as a `dict`, which is unchanged.

In practice, both of these actually occur while `_create_main_csv_headers` is called (i.e. before `_process_data_channel`), but the nature of the errors would be the same in either place.

moonraker595 (approved):
👍

PR #159, opened 2025-02-19
Implement YYYY/MM/DD/HHMMSS subdirectory structure

patrick-austin:
- New images/waveforms created in Echo will use a `YYYY/MM/DD/HHMMSS` (sub)directory structure, to avoid the current situation of having ~100,000 objects in the `image` and `waveform` directories, which made listing practically impossible when e.g. debugging with `s3cmd`.
- Minor name changes/refactors to some existing methods which extracted channel_name from the image/waveform_path so that they were more consistent and intuitively named across both classes.
- In `get_image` and `get_waveform`, first try the new subdirectory structure and if that fails, default to the old structure before actually failing. This allows us to use the new structure going forward, without needing to migrate existing data.
- Minor updates to relevant tests (coverage might need checking - locally I had a mix of "old" and "new" paths in my bucket but if the CI ingests from scratch it will only have "new" paths, so coverage might be lower)
- Added a basic script with will search Mongo, find "old" data paths, copy the waveform/image to its new location, update Mongo and delete the old object. When I tested this locally however, it took the best part of a minute to move one record. Kevin has said their are around 150,000 records on the dev server. It might be that by chance my record had a lot of images, but if every record took as long as that it could be 100 days to migrate the data to the new structure. Ideally I think it would be better to have the backwards compatibility for the old path, which we can deprecate at a later date if needed then try to migrate all the historic data. But the script is included for completeness.

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -8,7 +8,7 @@ application-import-names = operationsgateway_api,test
 import-order-style = google
 enable-extensions=G
 per-file-ignores =
-  test/*: S101, S303
+  test/*: S101, S303, F401, F811
   test/experiments/scheduler_mocking/models.py: N815
   operationsgateway_api/src/models.py: B902
   operationsgateway_api/src/records/waveform.py: E402
--- a/operationsgateway_api/src/records/echo_interface.py
+++ b/operationsgateway_api/src/records/echo_interface.py
@@ -58,6 +58,23 @@ class EchoInterface:
             )
             raise EchoS3Error("Bucket for object storage cannot be found")
 
+    @staticmethod
+    def format_record_id(record_id: str, use_subdirectories: bool = True) -> str:
+        """
+        Historically, objects had their record_id (YYYYMMDDHHMMSS) as a directory. This
+        can lead to large (~100,000) numbers of directories in the same bucket which
+        take too long to list when debugging.
+
+        By splitting the directories into the format YYYY/MM/DD/HHMMSS, the hope is to
+        improve findability. However since migrating large amounts of data can be
+        costly, it may be practical to support both the new and old formats when
+        querying.
+        """
+        if use_subdirectories and len(record_id) > 8:
+            return f"{record_id[:4]}/{record_id[4:6]}/{record_id[6:8]}/{record_id[8:]}"
+        else:
+            return record_id
+
     def download_file_object(self, object_path: str) -> BytesIO:
         """
         Download an object from S3 using `download_fileobj()` and return a BytesIO
--- a/operationsgateway_api/src/records/export_handler.py
+++ b/operationsgateway_api/src/records/export_handler.py
@@ -317,9 +317,7 @@ class ExportHandler:
                 if channel_name in self.function_types:
                     waveform_model = channel["data"]
                 else:
-                    waveform_model = Waveform.get_waveform(
-                        Waveform.get_relative_path(record_id, channel_name),
-                    )
+                    waveform_model = Waveform.get_waveform(record_id, channel_name)
             except Exception:
                 self.errors_file_in_memory.write(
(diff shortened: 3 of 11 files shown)


moonraker595 (commented):
Hi, overall it looks good. I just had a few comments:

I've tested it by checking out the branch and re-running the ingest script with wipe data enabled.

This puts data in the following format:

```bash
% s3cmd ls s3://og-alex-dev/images/
		DIR  s3://og-alex-dev/images/2023/
```

I've then checked out an older branch and re-ran the ingest script, again with the wipe data enabled, to format the data as it was, this gives:

```bash
% s3cmd ls s3://og-alex-dev/images/
		DIR  s3://og-alex-dev/images/20230605080000/
		DIR  s3://og-alex-dev/images/20230605090000/
		DIR  s3://og-alex-dev/images/20230605100000/
		DIR  s3://og-alex-dev/images/20230605110000/
		DIR  s3://og-alex-dev/images/20230605120000/
		DIR  s3://og-alex-dev/images/20230605130000/
		DIR  s3://og-alex-dev/images/20230605140000/
		DIR  s3://og-alex-dev/images/20230605150000/
		DIR  s3://og-alex-dev/images/20230605160000/
		DIR  s3://og-alex-dev/images/20230606120000/
```

I wanted to simulate running the ingest against existing echo data in the old format. I've then checked out the new branch (this one) and ingested an HDF file manually, as expected, this gives:

```bash
% s3cmd ls s3://og-alex-dev/images/
		DIR  s3://og-alex-dev/images/20230605080000/
		DIR  s3://og-alex-dev/images/20230605090000/
		DIR  s3://og-alex-dev/images/20230605100000/
		DIR  s3://og-alex-dev/images/20230605110000/
		DIR  s3://og-alex-dev/images/20230605120000/
		DIR  s3://og-alex-dev/images/20230605130000/
		DIR  s3://og-alex-dev/images/20230605140000/
		DIR  s3://og-alex-dev/images/20230605150000/
		DIR  s3://og-alex-dev/images/20230605160000/
		DIR  s3://og-alex-dev/images/20230606120000/
		DIR  s3://og-alex-dev/images/2025/
```

Should we be putting the test data, through the ingest script, in the `s3://og-alex-dev/images/2023/` folder? or should we keep it in the old format and add a new 2025 hdf file to simulate the new structure, this way it'll more accurately match what's in prod, which will be a mixture of both? 

Getting images in the new and old format works fine. We should test that we can get images out of the new structure and the old one using the same interface. So testing that we can get an image from`20230606120000` but also `20250228113756` (or whatever new test data we put in 2025 folder).

As discussed, it'd be handy to have a comment in the new script that you've created, just for context.

As a side note: i wrote a small app [here](https://github.com/ral-facilities/operationsgateway-hdf5-generator) to create new hdf files. This was handy for creating new files for testing the bit-depth stuff. All of the hdf files i had were already in the system, it may be handy, may be not.

moonraker595:
Apologies, I did mean pre-prod but the same principle applies. 

I'm concerned that:
1. If we have any issues on pre-prod relating to echo when EPAC start using it, we won't be able to easily replicate the folder structure that exists on pre-pro. I guess we could copy some data in the old path style manually if it is ever a problem. I appreciate that for Production, the old style isn't applicable.  
2. If we're supporting both, we should properly test for both. This would mean creating both styles in the test data, adding some new hdf5 files in the new format and increasing the associated test that checks for the number of files etc etc. However, as this doesn't apply to Production, we'd be creating tests and test data just for pre-prod which feels a bit of a waste of time.

Having thought about it a bit more, I'm happy to ignore this on the understanding that once EPAC start using pre-prod in anger we can essentially remove support for the old style as you mentioned and hopefully delete the simulated data from the pre-prod echo so It matches the desired production format.

moonraker595:
So unit tests are fine. Nice work 👍

PR #160, opened 2025-02-21
Support for vectors and float images

patrick-austin:
### Float images
- **Added `E704` to `.flake8` `ignore`.** `black` formats `...` for `@abstractmethod` onto one line, which violates this flake8 rule so ignoring it
- **Added `float_images` section to config.** Mostly similar to the `images` section, but with distinct values in case these should be treated differently (will definitely need different colourmaps).
- **Added `FloatImageModel`, `FloatImageChannelMetadataModel`, `FloatImageChannelModel`.** Mostly similar to the Image models, but with less metadata fields
- **Replace `PartialImportCheck._is_image_or_waveform_stored` with new `EchoInterface.head_object`.** Doing a head operation is more efficient than downloading the entire object just to check if it exists or not. Head will also generate a 404, but will be quicker to return if the object does exist.
- **Add `ExportHandler._add_float_image_to_zip`.** Support for exporting float images in their storage format, `.npz`. Chosen because "Stephen says that perhaps an export of a numpy array might be better because users will probably just want to read the data with Python." Also note that since "These are always displayed in false colour", so there isn't the need to return the "original" greyscale image like the normal images. npz was smaller than tiff when I tested them.
- **Added methods to apply false colour to float images.**
- **Introduced `FloatImage` controller class, and refactored common functionality to a `ImageABC`.**
- **Add channel checks for float images.**
- **Refactored repeated code in `ChannelChecks` into static utility functions.** Also pluralised the names for e.g. `ingested_images` to make it clearer they're lists.
- **Added `HDFDataHandler._extract_float_image`**
- **Refactored logic into `channel_checks` to avoid complexity limits from flake8.**

### Vectors
- **Add vector thumbnail size to config**
- **Add error for when Vectors can't be retrieved**
- **Added Pydantic models for Vectors**
- **Added export support for vectors**
- **Added ingestion and channel check support for vectors**
- **Added functions to insert, get, and generate .pngs from Vectors (as a bar chart)**

## Tests
Integration tests are running against a new record, `20230605080300`, which has been generated with v1.2 of the data sim and so includes channels for float images and vectors. ~This, and the updated channel manifest, have been pushed to the CI bucket.~ REVERTED. Current plan is to only push these once we are almost ready to merge, then tests can be re-run in order to limit CI failures here and for the frontend.

Note that as well as new tests/test cases, existing tests for get_records also needed updating (as they limit 2 and start at 0800, so 0803 is now record 2 in place of record 0900. But this new record will not affect the channel summary, since it was not in the last 3 records.

Can increase test coverage in a later PR which uses the CI generated report(s).

## Postman
- Get records
  - Added condition for the new record
  - Added projection for a float image channel
  - Added projection for a vector channel
- Get float image
- Get vector


moonraker595:
As discussed, the review is on hold until we get an updated EPAC-data-sim to more accurately test before pushing to main.

moonraker595 (commented):
Hi, looks good. The code seems logical, I've run it up locally as well and I can see all the tests pass. I think going forward, we should have some kind of a review meeting, where we go through these larger PRs. At `+2591, -616` changes, this is too big for me to meaningfully review without any context (I mean, what is a wavefront!?), particularly without any historical knowledge or JIRA ticket to provide some background. I don't think there's any getting away from these bigger PR's, but it'd certainly help my understanding going forward, mainly around the "why" aspect of the code change. I'll reach out on Slack for this one. 👍

moonraker595:
Just a note to say we had a face-to-face review. 
All looks good 👍

moonraker595 (approved):

PR #161, opened 2025-02-24
Updated Readme

moonraker595:
The Jira task required updating the README to include instructions for creating test data. However, while following the current instructions from a blank Rocky 9 VM, I noticed some key instructions were missing, which led me to rewrite it / shuffle some bits about to make it easier for someone to get the API up and running.

**Side note:** In the process of doing this, I discovered that the latest version of Poetry isn't compatible with our noxfile.py due to the [export function](https://python-poetry.org/docs/cli/#export) being removed from the core install. Updating the poetry dependencies meant doing a `poetry install` which updated Flake to a version which includes more stringent checks. At this point, i thought this relatively simple PR was ballooning so I have created another [task in JIRA to capture this](https://stfc.atlassian.net/browse/DSEGOG-383).


README.md
```diff
@@ -4,55 +4,146 @@
...
-- `safety` - this uses [safety](https://github.com/pyupio/safety) to check the dependencies (pulled directly from Poetry) for any known vulnerabilities. This session gives the output in a full ASCII style report.
-- `tests` - this uses [pytest](https://docs.pytest.org/en/stable/) to execute the automated tests in `test/`. There are currently no automated tests written for this repository however.
+Add:
 
-## API Configuration
-In `operationsgateway_api/`, there is an example config file (`config.yml.example`). Copy this example so `config.yml` exists in the same directory level and edit the configuration as needed for your system. There is also a logging configuration file (`logging.ini.example`). Follow the same procedure as the config file as to set up the logging.
+```bash
+[mongodb-org-7.0]
+name=MongoDB Repository
+baseurl=https://repo.mongodb.org/yum/redhat/8/mongodb-org/7.0/x86_64/
```
patrick-austin, line 63: Given the instructions are intended to be tailored for Rocky9, should this use https://repo.mongodb.org/yum/redhat/9/ instead of https://repo.mongodb.org/yum/redhat/8/? My dev VM is Rocky8 so mine does match this but I'd have thought it would be different to work on 9?

README.md
```diff
@@ -4,55 +4,146 @@
...
+```bash
+poetry run python util/realistic_data/ingest_echo_data.py
+```
+
+More details about this script can be found in `docs/developer/test_data.md`
+
+## Starting the API
+
+```bash
+poetry run python operationsgateway_api/src/main.py
```
patrick-austin, line 118: I think Viktor mentioned to me that the current recommended approach is to use [`fastapi run`](https://fastapi.tiangolo.com/deployment/manually/?h=run#use-the-fastapi-run-command), so for us:
```bash
poetry run fastapi run operationsgateway_api/src/main.py
```
Having tried it just now I get more logging and some fancy formatting, but it stops early because my code has syntax errors as I'm mid-development - so I'm not 100% on what would change if we used it, but it might be worth considering?

patrick-austin (commented):
Most of it looks fine to me - I just noticed that the mongo url still has rhel 8 in it, and wondered about using `fastapi run` instead of `python`.

Neither of those is critical, but I can't fully test as I'm mid-development on a Rocky8 machine. I'm happy to either keep them as they are if the link is correct and running with `python` is preferable, but thought I'd flag them just in case.

README.md
moonraker595: Good spot. I feel this needs investigating a bit more to fully understand what it does. I've created an issue [here](https://stfc.atlassian.net/browse/DSEGOG-385)

README.md
moonraker595: They must be cross-compatible as I've been using redhat/8 for a while, however, updating to 9 works fine as well. I have updated the readme.

README.md
patrick-austin: Yeah I suppose it might also be of interest in the ansible repo, I don't know how that actually runs things for deployment. There's also `fastapi dev` which does the hot reloads when the source code changes.

patrick-austin (approved):

PR #163, opened 2025-02-26
Support for raw files

patrick-austin:
## Changes
- **Added presigned url lifetime to echo config.** Used when generating presigned urls (download links) for the raw file channels as part of a records request, with the intention (assumption) that this could be displayed in the frontend as a clickable link/icon in table view.
- **Added exceptions for when raw files not found.**
- **Added models for raw files.** In addition to the raw bytes and Echo storage path, the only other metadata specified in the OpsGateway data structure document is `original_filename`, though it does not give intended us of this. Potentially could be used in the frontend table view?
- **Added `EchoInterface.generate_presigned_url`**
- **Added `ExportHandler._add_raw_file_to_zip`**
- **Added channel checks for raw files**
- **Added HDFDataHandler._extract_raw_file**
- **Added `RawFile`.**

Other changes are shared with those proposed in #160 (head_object, refactors, etc.)

## Tests
As with #160, until the data sim is updated to produce raw file / byte data, our testing model makes it impossible to get coverage without appropriate channels (and there is also the possibility that rather than being a bytes string, the raw file may be hard linked as this is a feature of hdf5 files).


moonraker595:
As discussed, the review is on hold until we get an updated EPAC-data-sim to more accurately test before pushing to main.

patrick-austin:
Closing as no longer a priority for EPAC

PR #165, opened 2025-03-03
User prefs for skip/limit vector thumbnails

patrick-austin:
## Changes
- **Add method for that doesn't error when user pref not found but instead returns default**
- **Add names for user prefs for limiting the display of vectors**
- **When applying false colour to thumnails, also re-generate the vector thumbnails if the user has truthy values for skip or limit**
- **Optionally crop data when generating vector thumbnails**


moonraker595 (commented):
Looks good, i was wondering what the defaults were, but by the looks of it they're set to `None` if they're not in the db. Could/should the skip/limit defaults be set in the config?

moonraker595 (approved):

PR #171, opened 2025-03-13
Update delete_by_record_id to use subdirectories

patrick-austin:
While discussing the deletion of old records in the sprint, I noticed that the `delete_record_by_id` endpoint was using the record_id directly in the Echo path. This is fine for the historical data, but since the new data goes in subdirectories by default we should try and purge both locations.


moonraker595 (approved):
Hi, Looks good. I have spun it up locally and can add , get and remove a file from the new style directory structure.

I've also checked it against the old style format as well, by checking out tag `v0.1` which didn't have the new dir structure, and rerunning the ingest script to force ingestion in the old style format.

PR #172, opened 2025-03-20
Adding healthcheck user

moonraker595:
To check that the API is working, a script needs to be written that makes the appropriate calls and outputs the result in a format Icinga can understand. 

The API "working" is defined as being able to `/GET`, `/PUT`, `/DELETE` data from Mongo and echo. This will likely be expanded in the future to accommodate the checking of the `background_scheduler` cron job.

The python script itself will be deployed on the rig machine using the Ansible script. 

The ansible script manages credentials by storing them in a vault and then replacing the necessary auth item in the [config](https://github.com/ral-facilities/operationsgateway-ansible/blob/main/roles/operationsgateway-api/templates/config.yml.j2).
In order to keep all credentials in the same place, this PR:
-  creates the necessary user in the config file,  for the healthcheck script to grab from the file
-  creates the necessary test user with appropriate route permissions

This PR has a matching one in the ansible repo [here](https://github.com/ral-facilities/operationsgateway-ansible/pull/28).

I haven't included any automated tests as I'd class this as a deployment thing rather than a functional API change, but let me know if you feel strongly about it either way.

### DIFF
--- a/operationsgateway_api/config.yml.example
+++ b/operationsgateway_api/config.yml.example
@@ -9,7 +9,7 @@ app:
   maintenance_file: operationsgateway_api/maintenance.json.example
   scheduled_maintenance_file: operationsgateway_api/scheduled_maintenance.json.example
 healthcheck:
-# Defines the username & password needed for a script that checks the api is working
+  # Defines the username & password needed for a script that checks the api is working
   username: username
   password: password
 images:
--- a/operationsgateway_api/config.yml.example
+++ b/operationsgateway_api/config.yml.example
@@ -8,6 +8,10 @@ app:
   url_prefix: ""
   maintenance_file: operationsgateway_api/maintenance.json.example
   scheduled_maintenance_file: operationsgateway_api/scheduled_maintenance.json.example
+healthcheck:
+# Defines the username & password needed for a script that checks the api is working
+  username: username
+  password: password
 images:
   # Thumbnail sizes should only ever be two element lists, of a x, y resolution
   thumbnail_size: [50, 50]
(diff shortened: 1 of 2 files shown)


patrick-austin (commented):
I'm happy for this without dedicated tests, if we wanted to have automated tests of the healthcheck script then we could move the script here (i.e. if we want to test it as source code it should live with the source code, if we don't then it's best for it to live in the ansible repo).

Like my comments on the ansible (i.e. that the creation of the user is done manually), I just want to check the additional line in `user_for_mongoimport.json` is for example purposes rather than actually being used by the ansible (as I couldn't see it being used anywhere).

moonraker595:
Hi, yes you're correct, the additional line in `user_for_mongoimport.json` is for example purposes. 

As a complete side note, but somewhat relevant; I'm not actually able to automatically import the users from `user_for_mongoimport.json` and I end up having the `mongosh` into my local instance and add the user manually. So [this bit](https://github.com/ral-facilities/operationsgateway-api/blob/abf3da01b52999a2b31fe281f9de7b10a706fd42/util/realistic_data/ingest_echo_data.py#L44) doesn't work for me. I presume its probably down to my set up, so not a massive issue.

patrick-austin (approved):
> As a side note, but somewhat relevant; I'm not actually able to automatically import the users from user_for_mongoimport.json when I rerun the ingest script, and I end up having to mongosh into my local instance and add the user manually. So [this bit](https://github.com/ral-facilities/operationsgateway-api/blob/abf3da01b52999a2b31fe281f9de7b10a706fd42/util/realistic_data/ingest_echo_data.py#L44) doesn't work for me. I presume its probably down to my set up, so not a massive issue.

I have had a problem in the past (probably when doing the maintenance endpoints since that involved altering the JSON file) but can't remember the details, e.g. whether it was intermittent or something I was able to ignore/work around. If I run into trouble again I'll try and look into it more closely, in case it is an issue to be fixed.

PR #173, opened 2025-03-26
Increase wavefront channel test coverage and bugfixes

patrick-austin:
Increases coverage for lines missed by the tests in #160 

Also has a few bugfixes that came up:
- Use `PartialVectorChannelModel` (this was preventing Vector metadata from being returned by requests to /records)
- `_check_type` returns the attribute, so that `_check_list` can actually function as intended
- Some changes to the messages of the various metadata checks. I'm not 100% confident about these, but basically a lot of the messages generated do not match the list at the end of the functions. I think they should, as otherwise those errors won't be propagated? I didn't have time to properly test this, so it might have been OK as is?
- Replace use of keys with attributes in `_process_data_channel` (probably a hangover from #158 )


moonraker595 (approved):
Looks good. There are a few places that use the same HDF file creation set-up:
```
with h5py.File("test.h5", "w") as f:
    f.attrs.create("epac_ops_data_version", "1.2")
    record = f["/"]
    record.attrs.create("timestamp", "2020-04-07 14:28:16")
    record.attrs.create("shotnum", 366272, dtype="u8")
    record.attrs.create("active_area", "ea1")
    record.attrs.create("active_experiment", "90097341")

Just wondering if this could be abstracted out, or `create_test_hdf.py` could be used?

This is fairly minor, so I'll leave it up to you.

PR #174, opened 2025-03-26
Adding version endpoint to get hash from config file

moonraker595:
Added a simple endpoint to display the deployed version of the API. 

As we start to look at a production instance, a version endpoint would be useful to be able to differentiate between prod and pre-prod versions, instead of relying on recently implemented functionality. 

A config item should be added with the version set to “test version”. Since we use the git hash in Ansible to deploy the correct version, it's the ansible that will set the hash version in the config.version property, as it does for other values.

Todo:

- [x] Code and manual tests
- [x] Automated tests
- [x] Update postman
- [x] Update Anisble to fill in hash version, See PR https://github.com/ral-facilities/operationsgateway-ansible/pull/29


moonraker595:
Thanks for this, i wasn't aware we were storing the version in the toml. I'll pick this up tomorrow.

moonraker595:
Looking into this more, we should absolutely be displaying what's in the `pyproject.toml`, it makes the process of displaying a version a lot simpler and less hidden.

Firstly, I think we should have a separate deployment process as I don't feel comfortable automating the Ansible deployment just yet. But as you've mentioned, we can give Ansible a version tag to deploy and I'll try that out during the next deployment.

This is what I've been looking at for the API:
- Updating the version upon creating a tag.
   I like using the UI to create tagged versions. You can set it to auto-populate the release notes, but you can add any context relevant to the release. So I thought about an action that would pick up the tag, update the version in the pyproject.toml and commit that back to main. However, that wouldn't work because the tagged version would still contain the old version in the pyproject.toml and there'd be a mismatch between main and the tagged version.
- Creating an action that picks up an update to the pyproject.toml and automatically creates the tag with whatever is specified in the version field. However, this would mean automatically creating releases with either none or the "standard" release notes.
- Re-creating the tag:
   Say I create a release with tag v0.3 through the GitHub UI. GitHub Actions picks it up and:
Extracts the version from the tag (0.3)
Bumps pyproject.toml to 0.3
Commits the version bump
Creates a new tag (v0.3.0-fixed or v0.3.0-final) pointing to the updated commit
The Ansible then deploys the updated tag (e.g. v0.3-final).
However, this would mean duplicating releases which is not ideal.

I think a lot of these problems come from me wanting to use the UI to create the tags. There is a GitHub CLI command to auto-generate the release notes: https://cli.github.com/manual/gh_release_create, so I'll investigate that.

patrick-austin:
> I like using the UI to create tagged versions. You can set it to auto-populate the release notes, but you can add any context relevant to the release. So I thought about an action that would pick up the tag, update the version in the pyproject.toml and commit that back to main. However, that wouldn't work because the tagged version would still contain the old version in the pyproject.toml and there'd be a mismatch between main and the tagged version.

You could force update the tag, we do this for one of the software packages as part of the Galaxy project, where we create a tag, triggers a CI build, create a DOI on Zenodo for that version of the software, then want to include that DOI in the docs for the package. I can't remember the commands (and I don't think it's automated as part of the CI) but if we can configure it to bump the version in the `toml` I think forcing the tag wouldn't be any harder.

As a low tech option, we could just try to keep the `toml` with the next anticipated release version. E.g. after releasing 0.3.0, we set it to 0.4.0 as part of the next change so that when it comes time to release, it doesn't need modifying. Downside is that you'd still need to tweak it manually if you wanted to do a patch release. This is a bit more similar to the ICAT Java components (the version is set to the next one with `-SNAPSHOT` on the end after release).

moonraker595:
Hi, Thanks for the suggestions. After taking the weekend to think about this, I'm going to leave the automatic version releasing for now.[ I've added something on the backlog](https://stfc.atlassian.net/browse/DSEGOG-416) to pick up (including automatic deployment) when I have some time to kill. With time ticking on and funding being a bit "ropey", i'd rather spend what time is left on fixing the bugs, metrics and future proofing the dependencies.

So for now, the release process is:

1. Update the version in `pyproject.toml`, which is reflected in the `/version` endpoint
2. Create a release with the same tag
3. Use that tag to deploy in Ansible

I've updated the docs to reflect this.

I've also updated the pyproject.toml for the anticipated new version and added us both as authors (in order of time spent on the project).

patrick-austin (approved):
That all sounds sensible to me, I think the three step process you describe is fine for the time being.

PR #175, opened 2025-04-08
DSEGOG-409 Better swagger docs for ingest endpoint

moonraker595:
I'm not sure what documentation Steven is expecting but updating the swagger docs for the endpoint seems like a good start. This PR adds a response model to the endpoint so the swagger docs are more verbose.

## Schema view
Screenshot 2025-04-08 at 19 20 35

## Example View
Screenshot 2025-04-08 at 19 21 02


operationsgateway_api/src/routes/ingest_data.py
```diff
@@ -32,6 +36,19 @@
     summary="Submit a HDF file for ingestion into MongoDB",
     response_description="ID of the record document that has been inserted/updated",
     tags=["Ingestion"],
+    response_model=SubmitHDFResponse,
+    responses={
+        201: {
+            "description": "Created and inserted new record with warning",
```
patrick-austin, line 42:
```suggestion
            "model": SubmitHDFResponse,
            "description": "Created and inserted new record with warning",
```

Having looked at https://fastapi.tiangolo.com/advanced/additional-responses/#additional-response-with-model, it seems like you can explicitly identify the model used for the status code, which would let us get the schema for both the 201 and 200 case (in the screenshots it only shows for 200 - I assume it defaults to the lowest code when set by `response_model` on line 39?)

operationsgateway_api/src/routes/ingest_data.py
```diff
@@ -32,6 +36,19 @@
...
+            "content": {
+                "application/json": {"example": example_created_response_with_warning},
+            },
+        },
+        200: {
+            "description": "Updated existing record",
```
patrick-austin, line 48:
```suggestion
            "model": SubmitHDFResponse,
            "description": "Updated existing record",
```


Having looked at https://fastapi.tiangolo.com/advanced/additional-responses/#additional-response-with-model, it seems like you can explicitly identify the model used for the status code, which would let us get the schema for both the 201 and 200 case (in the screenshots it only shows for 200 - I assume it defaults to the lowest code when set by `response_model` on line 39?)

patrick-austin (changes requested):
Didn't know you could set an example response/class for each error code separately. I've used the Pydantic `examples` argument for fields before but this is a really good way of being able to give more detail for situations like this where the response codes are the same class but conceptually distinct.

One incredibly minor thing I did notice in the screenshots was the schema only appears for one code, but looking at the docs I think we can get it to display for both (this is with my suggested changes, locally):
image

patrick-austin (approved):
Test failure is caused by the merging of #176 and #160 , and is unrelated to this. I will fix as part of #173 (if it doesn't end up getting fixed by something else sooner).

PR #176, opened 2025-04-10
Dsegog 410 warning for file version number too high received as a char array instead of a string

moonraker595:
The warnings were being displayed as a list of chars instead of a list of strings.

Now the ingest endpoint returns, for example:

`"warnings": [
            "File minor version number too high (expected <=1)"
        ]`

I've had to suppress the linter for now and have [created an item in the backlog](https://stfc.atlassian.net/browse/DSEGOG-415) to deal with this later.

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -1,7 +1,7 @@
 # .flake8
 [flake8]
 select = A,B,B9,BLK,C,E,F,I,N,S,W
-ignore = E203,W503,E501,B905
+ignore = E203,W503,E501,B905,C901
 max-complexity = 12
 max-line-length = 80
 application-import-names = operationsgateway_api,test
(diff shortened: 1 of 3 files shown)


patrick-austin (approved):
Change itself is good.

Regarding C901, I ran into this a lot when working in #160 since I was adding another channel type in a lot of places. I did do a lot of refactors for common functionality (e.g. type checking of attributes) to get it down below the limit. So might be worth double checking if this still tips it over the edge once that's all merged as well.

moonraker595:
Thanks, I'll take a look. I've just added a test for checking the warning message. Appriciate you've just reviewed the code so let me know if you have any comments about the test code.

patrick-austin (approved):
Yeah test looks good
(I started reviewing yesterday before the CI finished so I could get it done before I left the office, didn't see the coverage until after I approved)

PR #177, opened 2025-04-14
Add documentation for wavefront channels implementation

patrick-austin:
Adds markdown developer facing documentation for wavefront channels.

Gives an overview of a bit of the theory behind what "wavefront" means in terms of what the users will be trying to interpret, details of the new channel types, and how they are used specifically for wavefront data.

Note that the theory is mostly remembered secondhand from what Stephen told us about the requirements a while ago, so I may have misremembered or filled in the gaps inaccurately. This isn't intended to be a complete description of the processes involved, but just to provide enough context/understanding for other devs to work around the new channels.

If I've not explained something, or have badly described something then I'm happy to tweak further so it's of most use to others.


moonraker595 (approved):
Apart from the fact that its crazy we can measure this kind if stuff!..this really helps understanding (mine at least) and I'm sure this will help if I, or anyone else, has to do wavefront stuff in the future. Thanks 👍

PR #179, opened 2025-04-30
Dsegog 412 matching shot numbers

moonraker595:
## Shotnum

After some back and forth over email with SD, he wants to make the shot number unique in the DB. 
The following test cases were agreed upon:

| Test Case | System State                                                                 | Test                                                                                     | Result | Notes                                                             |
|-----------|--------------------------------------------------------------------------------|------------------------------------------------------------------------------------------|--------|-------------------------------------------------------------------|
| 1         | N/A                                                                            | A HDF file is submitted with a unique Timestamp and no shot number                       | ✅     | Accept, as shot number is optional (201)                         |
| 2         | N/A                                                                            | A HDF file is submitted with no Timestamp and no shot number                             | ❌     | Reject, timestamp is required                                    |
| 3         | N/A                                                                            | A HDF file is submitted with no Timestamp and a unique shot number                       | ❌     | Reject, timestamp is required                                    |
| 4         | N/A                                                                            | A HDF file is submitted with no Timestamp and an existing shot number                    | ❌     | Reject, timestamp is required                                    |
| 5         | N/A                                                                            | A HDF file is submitted with a unique Timestamp and a unique shot number                 | ✅     | Accept (201)                                                     |
| 6         | A record already exists with the same shot number                              | A HDF file is submitted with a unique Timestamp and an existing shot number              | ❌     | Reject, shot number must be unique – “shot number already exists”|
| 7         | A record exists with the ame timestamp and no shot number (`None`)               | A HDF file is submitted with the same Timestamp and no shot number                       | ✅     | Accept (200), treated as update                                  |
| 8         | A record exists with the same timestamp and a shot number                          | A HDF file is submitted with the same Timestamp and different shot number (e.g. 25145)       | ❌     | Reject, inconsistent metadata                                    |
| 9         | A record exists with the same timestamp and a shot number                          | A HDF file is submitted with the same Timestamp and shot number as `None`                    | ❌     | Reject, inconsistent metadata                                    |
| 10        | A record exists with the same timestamp and same shot number                       | A HDF file is submitted with the same Timestamp and the same shot number                         | ✅     | Accept (200), merge behaviour as currently implemented           |
| 11        | A record exists with same timestamp and the same shot number with a data version of 1.0 | A HDF file is submitted with the same Timestamp and the same shot number but a different version number (1.2) | ✅     | Accept (200), value must be present, but versions can differ. The version should not be updated in the DB    |

## Data version
Test 11 also does some checking around the data version. SD was keen for the API NOT to check the version. I have removed the checks and stopped the API from updating the version, as updating metadata isn't a thing yet.

## Updates to codebase
- `partial_import_checks.py`: I've refactored this to be less clunky, now we just check what we need to and reject anything else
- `record.py`: since we need to check for shotnum now, added a function to check for it as well as excluding version from the update
- `ingest_data.py`: previously, we only returned a record if the timestamp matched, which wouldn't work if, for example, the timestamp was new but the shotnum wasn't.
- `test_submit_hdf.py`: removed this test as it's covered in the new tests
- `create_test_hdf.py`: expanded so can use missing, default or custom shotnum
- `test_partial_import.py`: The tests matching the table above (EDIT: I tried using `parametrize` for these tests but was struggling to generalise them enough, plus I think being more explicit makes it more obvious whats going on, so stuck with that)


moonraker595:
I forgot to mention, this encompasses [JIRA task 418](https://stfc.atlassian.net/browse/DSEGOG-418)

patrick-austin (approved):
Looks OK to me, I agree that the explicit tests for the different scenarios probably makes more sense here (potentially if they were all failure state you can parametrise the input and string for the error message, but that wouldn't work here since some a success, some throw and some return a 400.

There's a conflict, presumably from the PRs I was merging the other day but it looks like it's just the imports for the test file so shouldn't be too bad.

PR #180, opened 2025-05-06
Add wavefront channel types to /channels/summary

patrick-austin:
The new wavefront types (`FLOAT_IMAGE` and `VECTOR`) were missing from the if/elif block for requests to `/channels/summary`. Have added these, along with a warning if the channel dtype is unrecognised. I considered making it throw an error (since the dtype should be well controlled) but the existing tests assert that this is an empty list so I left the functionality as is.


moonraker595 (approved):
Looks good, yeah not sure why `test_incorrect_channel_dtype` checks for an empty list. Leaving it as it is is fine though.

PR #184, opened 2025-06-12
Dsegog 414 investigate open search for metrics

moonraker595:
This PR adds the necessary code changes for the application to speak to the local APM server (to be set up on the machine, either manually or by the Ansible script). Ideally, we'd use environment variables for this, but I've stuck with the convention of having everything in a config file so as not to mix and match.

There might be a bit of back and forth with this as I try to get everything talking. First step is these changes, then to get the APM server working on the rig VM, then get the Elastic session set up. Mehmet has sent through instructions for these.

### DIFF
--- a/operationsgateway_api/config.yml.example
+++ b/operationsgateway_api/config.yml.example
@@ -78,3 +78,7 @@ export:
   # 1048576 = 1 MB
   # 1073741824 = 1 GB
   max_filesize_bytes: 1073741824
+observability:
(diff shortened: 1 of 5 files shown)


moonraker595:
@patrick-austin : I've just noticed a warning on startup that needs further investigation. The CI is also failing presumably due to the echo issue, so I'll remove the review for now until both are sorted.

PR #187, opened 2025-06-18
Elastic integration

moonraker595:
This PR makes the necessary changes to the code to be able to integrate with the Elastic agent, which will be installed on the dev server. 

The general plan is to:
- Get this code on the dev server
- Install the agent as per Mehmets instructions
- Try it out by logging into their es stack. 
- If everything works, update the Ansible to deploy the new creds and potentially auto-install the agent

If it doesn't work...roll back these changes

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -750,6 +750,48 @@ pipenv = ["pipenv"]
 poetry = ["poetry"]
 
 [[package]]
+name = "ecs-logging"
+version = "2.2.0"
+description = "Logging formatters for ECS (Elastic Common Schema) in Python"
+optional = false
+python-versions = ">=3.6"
+groups = ["main"]
+files = [
+    {file = "ecs_logging-2.2.0-py3-none-any.whl", hash = "sha256:f6e22d267770b06f797076f49b5fcc9d97108b22f452f5f9ed4b5367b1e61b5b"},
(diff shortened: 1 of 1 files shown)


moonraker595:
Actually, I might be easier if I deploy this branch into pre-prod, instead of merging into main?

operationsgateway_api/src/config.py
```diff
@@ -115,6 +115,13 @@ class ExportConfig(BaseModel):
     max_filesize_bytes: StrictInt
 
 
+class ObservabilityConfig(BaseModel):
+    """Configuration model class to store export observability details"""
+
+    environment: StrictStr
+    secret_key: StrictStr  # apm key
```
patrick-austin, line 122: Could/should this be a `SecretStr` https://docs.pydantic.dev/2.0/usage/types/secrets/ ?

patrick-austin (approved):
Change itself looks good, my only thought is whether we should be using Pydantic's `SecretStr` for things like keys or passwords.

There are a few other places like the Echo config or the experiments config that might also be storing sensitive values, so I'm happy to approve this and if we decide to use `SecretStr` then that should be a separate change and applied to all the relevant places (rather than doing it ad hoc).

operationsgateway_api/src/config.py
moonraker595: Thanks for flagging this: I've created a new issue in the backlog: https://stfc.atlassian.net/browse/DSEGOG-432

PR #186, opened 2025-06-17
Updating dependencies

moonraker595:
Updates all dependencies (`poetry update`) and solves Pr's 183 & 182

PR #189, opened 2025-06-19
Return system defaults from get_preferred_colourmap and get_preferred_float_colourmap

patrick-austin:
While ingesting wavefront data to the dev instance, noticed that the thumbnail for the float image uses `viridis`:
image

Despite the fact that the "system default" for floats should be `bwr`:
https://github.com/ral-facilities/operationsgateway-api/blob/8955ad50b418149a575222f738347497c219ef7f/operationsgateway_api/config.yml.example#L25

Having looked into it, this was a combination of not calling `FloatImage.get_preferred_colourmap()` in the records endpoints and the fact that if a user pref isn't found, both the method for traditional and floating point images were returning `None`. MatPlotLib defaults to `viridis` if `None` is passed:
```python
>>> import matplotlib.pyplot as plt
>>> plt.cm.get_cmap(None).name
'viridis'
```
Which is undetectable for the traditional images where the "system default" is `viridis` anyway, but is bad for floats which should always use a diverging colourmap to show the positive/negative regions.

Updated the phash in the tests from the hash of:
FE-204-NSS-WFS_before

to:
FE-204-NSS-WFSFalse

Also changed some of the places we were checking if the colourmap was None to use a single assignment using `or` instead. There shouldn't be any functional impact of this, but gets high patch coverage since technically that line always runs without being counted as a "partial" on the if statement. Happy to revert this though if people find it less readable than an explicit if.

### DIFF
--- a/operationsgateway_api/src/records/false_colour_handler.py
+++ b/operationsgateway_api/src/records/false_colour_handler.py
@@ -39,7 +39,7 @@ class FalseColourHandler:
             pref_name = FalseColourHandler.preferred_colour_map_pref_name
             return await UserPreferences.get(username, pref_name)
         except MissingAttributeError:
-            return None
+            return FalseColourHandler.default_colour_map_name
 
     @staticmethod
     async def get_preferred_float_colourmap(username: str) -> str:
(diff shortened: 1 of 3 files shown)


patrick-austin:
Yes I think it's just because this branched from `main` before #186 was merged. I'll merge this now but if safety still complains I'll have another proper look at it.

PR #190, opened 2025-06-25
Unique session names per user

moonraker595:
The frontend needs a way of making the session names unique for each user. For example; different users may have the "test" session but a user can't have 2 sessions that are named the same. 

More conversation here: https://stfc.atlassian.net/browse/DSEGOG-423.

This can be achieved by creating an index on the session collection:
`db.sessions.createIndex(
  { username: 1, name: 1 },
  { unique: true }
);`
and then catching the exception in the API on session POST.

Exception flow:
- Mongo raises a `DuplicateKeyError` on` insert_one`
- This raises a `DuplicateSessionError` defined `interface.py`
- This gets picked up by the function wrapper in `sessions.py`
- A 400 is returned with:`{
    "detail": "Session name already exists for this user."
}`

For testing, the calls have been mocked, but for pre-prod, prod and probably your local instance, the index will need to be applied.

Once this is merged, then the index can be created on pre-prod and prod.

### DIFF
--- a/README.md
+++ b/README.md
@@ -80,7 +80,11 @@ sudo systemctl start mongod
 sudo systemctl enable mongod
 ```
 ### MongoDB Indexes:
-The following Indexes are uses and need to be set up on local, dev & prod databases:
+The following Indexes are uses and need to be set up on local, dev & prod databases.
+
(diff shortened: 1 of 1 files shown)


moonraker595:
Forgot to mention, this has been tested locally against a MongoDB with the index applied

patrick-austin (approved):

PR #192, opened 2025-06-27
replacing some StrictStr with SecretStr

moonraker595:


patrick-austin (changes requested):
`util/ingest_hdf.py` uses values from the "main" config file:
https://github.com/ral-facilities/operationsgateway-api/blob/7243f9cae296e83a2d44c4c0056b91aac6e323bd/util/ingest_hdf.py#L16
https://github.com/ral-facilities/operationsgateway-api/blob/7243f9cae296e83a2d44c4c0056b91aac6e323bd/util/ingest_hdf.py#L122-L123

So as things stand I think these would need `.get_secret_value()` as well?

I think this raises some broader questions though:
1. https://github.com/ral-facilities/operationsgateway-api/blob/SecretStr/util/realistic_data/ingest/config.py also includes values which are probably sensitive, but are only used in the context of the "realistic_data" ingest scripts. Should this also be updated?
2. Should `ingest_hdf.py` use the values from the "main" config or the "realistic_data config"?
3. Tests seemingly do not cover these files in `util` (as otherwise I imagine we'd have had a test failure)? Is this a problem that should be address?

In the short term the easiest thing is probably to just add the `.get_secret_value()` where it's needed. I'm not 100% confident on when the various scripts in util are actually (supposed to be) run, and so what the impact of changes might be / whether it's all still needed.

PR #193, opened 2025-07-14
Write hdf to local directory on ingest to /submit/hdf

patrick-austin:
Adds config and functionality for writing the incoming hdf file to local storage, so that it can be copied to tape by a secondary process.

PR opened against 353 branch to keep the child issue PRs small without merging to main 1 by 1 as they are interdependent on shouldn't block us developing other functionality.

Anticipate using XRootD client for more functionality in the other child issues, but makes sense as a logical place to write this static function.

As there can be multiple hdf files for the same record, we need a way of tracking the order in which the hdf files were ingested so  introduce a new `version` field to the Record model. Largely this will have no  impact as requests from the frontend use projections to only request the fields they care about.


operationsgateway_api/src/models.py
```diff
@@ -200,6 +200,13 @@ class RecordMetadataModel(BaseModel):
 
 class RecordModel(BaseModel):
```
moonraker595, line 201: Do we need an example in the `config.example` for this?

moonraker595 (changes requested):
Hi, looks great. Just a quick question besides the config one: I’m presuming the script that writes this to tape will be responsible for deleting it from local disk? We’d also need some alerting around that script, it would be grim if it silently failed. Maybe email alerts? Note; I had trouble integrating a script.py with Icinga and never really got it working due to SElinux rules.

moonraker595:
> > I’m presuming the script that writes this to tape will be responsible for deleting it from local disk?
> 
> Yes, the current WIP has a callback with runs once the transfer to the tape buffer is complete. Initially I thought this might be a separately running job but it turns out the XRootD client library can support that sort of cleanup out of the box.
> 
> > We’d also need some alerting around that script, it would be grim if it silently failed. Maybe email alerts? Note; I had trouble integrating a script.py with Icinga and never really got it working due to SElinux rules.
> 
> Current WIP implementation is just logging at ERROR if it fails. I'm not sure with the new Elastic traces, observability etc. whether it's possible to extract the information we're already logging and escalate it? But if that's not possible/practical then I can add in a config option for a mail server to notify for unexpected failures.
> 
> Either way all the logic for this will be in another PR, possibly with some of the logic refactored into a common xrootd-utils" repository.

Hi, we'll need to speak to Mehmet about the Elastic observability option. I know it's possible, but I think it's a paid option. However, I got the impression he was after use cases to justify the purchase. But I agree this should be the preferred option, and if we can't work anything out, then a mail server is the way to go.... probably @segsupport

moonraker595:
> > Hi, we'll need to speak to Mehmet about the Elastic observability option. I know it's possible, but I think it's a paid option. However, I got the impression he was after use cases to justify the purchase. But I agree this should be the preferred option, and if we can't work anything out, then a mail server is the way to go.... probably @segsupport
> 
> Ah OK - I do have access to the (dev?) Elastic cluster and (think) I saw it in the side panel but didn't look into it beyond checking my credentials worked. Possibly this is a combination of wishful thinking/misremembering the presentation on observability from the group meeting and thinking we could piggyback an existing solution as long as everything was logged to stdout.

Sorry, just to clarify. I think the notification/email sending is an optional extra. We should get all the logging and traces for free. Anyway, have a play and see what's possible. 👍

moonraker595 (approved):
Looks good

PR #194, opened 2025-07-14
Add lifespan create job to put_bucket_lifecycle_configuration

patrick-austin:
Add the optional `PositiveInt` `echo.expiry_days`. If set, then use the FastAPI lifespan to, on startup, set this number of days for expiry of objects in the storage bucket.

Note that while doing this on startup allows you to update the policy with any change to `expiry_days` in the config file, if the value is removed entirely then the "old" lifecycle will still be in effect. We could either leave as is (relying on manually removing an old policy with `s3cmd` or similar) or have the FastAPI lifespan delete any in place lifecycle if `expiry_days` is `None`.


operationsgateway_api/src/main.py
```diff
@@ -71,7 +72,13 @@ async def get_experiments_on_startup():
         else:
             log.info("Scheduler background task has not been enabled")
 
+    @assign_event_to_single_worker()
+    async def set_bucket_expiry() -> None:
```
moonraker595, line 76: Does this need to async? `put_lifecycle` uses the sync boto lib.

operationsgateway_api/src/records/echo_interface.py
```diff
@@ -81,6 +81,30 @@ def format_record_id(record_id: str, use_subdirectories: bool = True) -> str:
...
+                "Prefix": "",
+                "Status": "Enabled",
+            }
+            try:
+                self.client.put_bucket_lifecycle_configuration(
+                    Bucket=Config.config.echo.bucket_name,
+                    LifecycleConfiguration={"Rules": [rule]},
+                )
+            except ClientError:
+                log.exception("Failed to put lifecyle")
```
moonraker595, line 104: We should probably log the bucket name here, and anywhere else has tries to set / delete bucket policy

moonraker595 (changes requested):
Hi looks good. Just a couple of comments. Its complaining about the coverage, is that because it's difficult to test?

operationsgateway_api/src/main.py
patrick-austin: Yeah I think you're right; I can't recall any reason it would need to so probably a copy paste from `get_experiments_on_startup` above.

Having said that, if/when #199 is merged then it might end up needing to change to async anyway. Depending on what order things get merged I'll sort it out in the conflict resolution one way or another.

patrick-austin:
> Hi looks good. Just a couple of comments. Its complaining about the coverage, is that because it's difficult to test?

Yeah so prior to this PR the `lifespan` function in `main.py` wasn't covered by tests, and this PR adds the bucket lifecycle stuff there so it also isn't covered:
https://github.com/ral-facilities/operationsgateway-api/blob/5e98317d201558f35a68349e3a3dcd3b3cf58b43/operationsgateway_api/src/main.py#L61-L84

I sort of justified this to myself as following the same testing pattern, and if the last person the write `lifespan` thought it was too hard to test, then I trust them.

Having said that, one of the changes introduced in #199 is:
> Replace TestClient(app) with with TestClient(app) to use the lifespan (see above)

So if/when that's merged then I _think_ we'll get coverage of the added lines. When it comes to the conflict resolution/merge I'll double check the coverage again.

patrick-austin:
> So if/when that's merged then I _think_ we'll get coverage of the added lines. When it comes to the conflict resolution/merge I'll double check the coverage again.

Having merged in the changes introduced by #199, the coverage on main should be higher. Looked like line 89 missing locally, so probably most of it gets executed by the tests but the way it's configured for tests means the `assign_to_single_worker` wrapped function isn't called?

PR #199, opened 2025-07-24
Implement lru_cache for EchoInterface

patrick-austin:
Will have conflicts with #195 as they both alter the get method for channel objects.

Replaces the `EchoInterface()` calls with `lru_cache` to prevent the overhead of initialising the object (up to ~0.4 seconds) for example in export which can call this 10s - 100s of times. In practice, the time it takes for all Echo calls is very inconsistent so the benefit may be less noticeable.

As a precaution against holding the same boto3 connection open for the lifetime of the application, during error handling for the calls which use `echo_interface` invalidate the cache. This means the next time we try to get the interface, it will be a fresh one.

### DIFF
--- a/operationsgateway_api/src/records/echo_interface.py
+++ b/operationsgateway_api/src/records/echo_interface.py
@@ -1,3 +1,4 @@
+from functools import lru_cache
 from io import BytesIO
 import logging
 
@@ -197,3 +198,12 @@ class EchoInterface:
                 f"{exc.response['Error']['Code']} when deleting file at"
                 f" '{dir_path}'",
             ) from exc
+
+
+@lru_cache
+def get_echo_interface() -> EchoInterface:
+    """
+    Returns:
+        EchoInterface: Cached object for interacting with Echo object storage.
+    """
+    return EchoInterface()
--- a/operationsgateway_api/src/records/float_image.py
+++ b/operationsgateway_api/src/records/float_image.py
@@ -11,7 +11,7 @@ from operationsgateway_api.src.auth.jwt_handler import JwtHandler
 from operationsgateway_api.src.config import Config
 from operationsgateway_api.src.exceptions import EchoS3Error
 from operationsgateway_api.src.models import FloatImageModel
-from operationsgateway_api.src.records.echo_interface import EchoInterface
+from operationsgateway_api.src.records.echo_interface import get_echo_interface
 from operationsgateway_api.src.records.false_colour_handler import FalseColourHandler
 from operationsgateway_api.src.records.image_abc import ImageABC
 from operationsgateway_api.src.records.thumbnail_handler import ThumbnailHandler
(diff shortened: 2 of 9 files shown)


moonraker595 (commented):
Hi, overall I'm happy with the quality of the code. I haven't got it up and running as I trust you have done the manual testing (as evident in [JIRA task](https://stfc.atlassian.net/browse/DSEGOG-447?focusedCommentId=104995)) but as we haven't got any concrete performance requirements, any improvement is good I guess!

I have a couple of questions/general observations:
- I feel like this is a bit of a halfway house. Clearly, you've done a lot of research on Python/Gunicorn's performance, which is impressive, but it massively complicates the code, which I worry would have an impact on maintenance down the line. But I get it... we have to do what we can with the “free” storage before we can make a case for the paid network-attached storage (which they are already paying for under the FDP project!... anyway rant over!).
- If we were to get rid of echo for a more performant solution, would this have an impact on the tape storage work you’re currently doing? Would we not get the back-ups done “for free” then?
- I think it’s also worth documenting the concurrency strategy in the developer docs somewhere, most of what you’ve put in the PRs can probably go in. 
- Is it safe (enough) to write to one zip file from multiple async tasks? From what I can gather, `TaskGroup` creates concurrent tasks which could ultimately all write to the same zip file at the same time? It doesn’t look too difficult to implement [Async locking ](https://docs.python.org/3/library/asyncio-sync.html)in Python, but this may have an impact on performance.

patrick-austin:
> * If we were to get rid of echo for a more performant solution, would this have an impact on the tape storage work you’re currently doing? Would we not get the back-ups done “for free” then?

No, not really. As far as I know nothing is backed up automatically as a matter of course from any of the storage offerings. While the purpose of Facilities Data Pipeline is to move data from online CephFS storage to tape, this is done based on http requests to archive specific datasets, creates metadata in ICAT, uses a more complicated transport mechanism etc. so  that also wouldn't help us here I don't think.
 
> * I think it’s also worth documenting the concurrency strategy in the developer docs somewhere, most of what you’ve put in the PRs can probably go in.

Yeah absolutely, if you're happy(ish) with the overall concept then I'll add some markdown on this.

> * Is it safe (enough) to write to one zip file from multiple async tasks? From what I can gather, `TaskGroup` creates concurrent tasks which could ultimately all write to the same zip file at the same time? It doesn’t look too difficult to implement [Async locking ](https://docs.python.org/3/library/asyncio-sync.html)in Python, but this may have an impact on performance.

Potentially there's a risk - I didn't notice anything while testing but I could've got lucky. I'll have a look and see if I can force an error to test this, and even if not I'll try adding an extra layer of defensive programming with the lock. Some of the asyncio stuff doesn't work out of the box because FastAPI/uvicorn is using the event loop (e.g. I don't think `asyncio.run()` works) but I can try this (and potentially alternatives if needed).

patrick-austin:
> Potentially there's a risk - I didn't notice anything while testing but I could've got lucky. I'll have a look and see if I can force an error to test this, and even if not I'll try adding an extra layer of defensive programming with the lock. Some of the asyncio stuff doesn't work out of the box because FastAPI/uvicorn is using the event loop (e.g. I don't think `asyncio.run()` works) but I can try this (and potentially alternatives if needed).

Just to follow up on this, I wasn't able to force an error by simultaneously writing. I messed about by adding some arbitrary waits in, but because `writestr` is synchronous I don't think the event loop will run it at the same time as any other synchronous code. Even though it's called from an async function, once the bytes from any of the concurrent co-routines come back from echo I think it runs synchronously until the next await, then it checks to see if any of the other co-routines have finished.

Having said that I'm not 100% confident, and having tested it there was no noticeable overhead from including the lock so I'm happy to include it as a defensive precaution (i.e. even if it seems/is safe today, maybe a future refactor makes it risky).

Have copied some of the high level implementation details to a markdown file along with some links and a general summary of how the TaskGroup stuff differs from the concurrency that the Uvicorn workers provide. A lot is from memory so there may be inaccuracies and always happy to add detail/rephrase to make it better for someone else reading it.

There's been vulnerability reported in `starlette`, requires a bump to the FastAPI version to fix to will open that separately.

moonraker595 (approved):
Thanks for adding the docs, all looks good. The CI is failing, I presume due to the safety issue addressed in the other PR.

PR #201, opened 2025-08-14
Wrap EchoInterface.download_file_object in alru_cache

patrick-austin:
Note: this is dependent on #199 as it uses `alru_cache`. You could do basically the same thing with `lru_cache`, but given Stephen's recent email requesting that we put the asynchronous/concurrent Echo onto prod I'm assuming that the former is the right thing to use.

- Add `cach_maxsize` config option, which defaults to the same default `maxsize` of `alru_cache` (128)
  - Based on looking at `htop` while the API runs, this seemed to result in ~1GB more RAM usage
- In order to use `alru_cache`, had to replace `BytesIO` with bytes as the return value of `EchoInterface.download_file_object`. `BytesIO` can be closed, which means it can't be safely cached and re-used like the raw `bytes`. In some places this now means we don't have to call `getvalue` to get the `bytes` since we already return them, in others (loading with PIL and numpy) we now wrap the `bytes` in `BytesIO`.
- Added a new test which wraps the underlying call and asserts that it is only called once if you request the same image twice (i.e. the second call uses the cached bytes and doesn't transfer anything from Echo)

### DIFF
--- a/operationsgateway_api/src/config.py
+++ b/operationsgateway_api/src/config.py
@@ -8,6 +8,7 @@ from pydantic import (
     BaseModel,
     field_validator,
     FilePath,
+    NonNegativeInt,
     SecretStr,
     StrictBool,
     StrictInt,
@@ -62,6 +63,7 @@ class EchoConfig(BaseModel):
     access_key: SecretStr
     secret_key: SecretStr
     bucket_name: StrictStr
+    cache_maxsize: NonNegativeInt = 128
 
 
 class MongoDB(BaseModel):
--- a/operationsgateway_api/src/records/channel_object_abc.py
+++ b/operationsgateway_api/src/records/channel_object_abc.py
@@ -1,5 +1,4 @@
 from abc import ABC, abstractmethod
-from io import BytesIO
 import logging
 
 from operationsgateway_api.src.exceptions import EchoS3Error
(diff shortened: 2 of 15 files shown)


operationsgateway_api/src/config.py
```diff
@@ -62,6 +63,7 @@ class EchoConfig(BaseModel):
     access_key: SecretStr
     secret_key: SecretStr
     bucket_name: StrictStr
+    cache_maxsize: NonNegativeInt = 128
```
moonraker595, line 66: Do we need to add this to the `config.yml.example`? Also might be worth adding a comment there as to what its used for.

moonraker595 (changes requested):
Hi, looks good! 

Apart from the config comment, I just have one general comment:

So the catch only stores previously downloaded data? What are the chances that a record gets updated and redownloaded? Would we then be serving outdated data?

operationsgateway_api/src/config.py
patrick-austin: Since it takes a default I was being lazy by not including it but being more explicit can't hurt. Especially since there's some nuance in the fact that it's "enabled" by default, and to "disable" you would need to set it to 0.

moonraker595 (approved):
Looks good

PR #205, opened 2025-09-03
Dsegog 435 new lifespan job to move local files to tape using xrdcp

patrick-austin:


patrick-austin:
Changes based on live testing with @moonraker595
* Merged `main` (specifically to get the latest config for OIDC changes)
* Removed version pin for `libuv`
* Add explanation of XRootD authentication to `README.md`
* Add inline comments to backup section of `config.yml.example`
* Added new config optional setting `keytab_file_path`, which if set will set the environment variable XRootD needs (as an alternative to requiring the user to set it in their session explicitly)
* Added logging support for XRootD library in `logging.ini.example`
* Fixed some places where the backup config wasn't actually optional (i.e. trying to access `Config.config.backup` with a `is not None` check
* Validation of environment variables (so API explicitly errors on startup if the keytab is incorrectly defined or permissioned.
* Fixed missing "%" from logging
* Defined `xrootd-utils` dependency using pypi
* Changes to tests to ensure coverage and handling optional config settings properly

Missing coverage is either partials or existing lines without coverage that have been refactored - all new functionality should be covered.

PR #207, opened 2025-09-16
Optionally send emails for warning or higher logs in BackupRunner

patrick-austin:
Adds optional config for a mailserver, and in the event of warning or higher logs (failed transfers, high disk usage, unexpected errors) if the config is defined send this to the configured recipients.

Currently lives as a `staticmethod` on `BackupRunner` since it's the only place that uses it, but could easily be refactored.

Unit test mocks the SMTP server but I've tested it locally with the STFC mailserver and an arbitrary message.

(Also slightly tweaked the permissions logging as I thought knowing the path would be useful as well)


moonraker595 (commented):
Hi, No issues with the code. I was just looking at the SMTP docs for `SMTP(host=Config.config.backup.mail.host)`. it says:

> If specified, `host` is the name of the remote host to which to connect. If specified, `port` specifies the port to which to connect. By default, smtplib. SMTP_PORT is used. 

You mentioned you've tested it so it clearly works, but I'm wondering if we could/should be more explicit with declaring the port and whether it matters?

We should also test this in pre-prod to check we can get real emails. I personally get the Icinga emails for OG but I don't know if that means I'll get these seg support ones as well.

moonraker595 (approved):

PR #249, opened 2026-04-13
Gemini Changes

moonraker595:
This PR:

1. Adds string as a new data type:
Introduces string as a supported data type.

2. Allows string for shotnum field:
Updates the shotnum field to accept both int and string values.
Maintains backward compatibility with existing numeric shot numbers.

3. Add an optional type field to shotnum:
Introduces an optional type attribute to the model for shotnum datatype.
If type is not provided, the system infers the type based on the value (str or int).

4. Support sub-second timestamp ingestion:
Extends ingestion logic to handle timestamps with sub-second precision.

5. Enhancements to range converter:
Updates range conversion logic to correctly handle mixed shotnum types (int and string).

6. Clean up:
Adds a new lock file for dependency updates (which is the bulk of the reported line changes)


test/sessions/test_user_session.py
```diff
@@ -6,7 +6,7 @@
 import pytest
 
 from operationsgateway_api.src.auth.jwt_handler import JwtHandler
-from operationsgateway_api.src.constants import DATA_DATETIME_FORMAT
+from operationsgateway_api.src.constants import DATA_DATETIME_FORMATS
```
patrick-austin:
```suggestion
from operationsgateway_api.src.constants import SESSION_DATETIME_FORMAT
```

This wouldn't affect functionality, but in `operationsgateway_api/src/routes/sessions.py` we now import `SESSION_DATETIME_FORMAT` instead of `DATA_DATETIME_FORMAT`. I think it makes sense to use an explicit variable name there (even if the actual value is the same as `DATA_DATETIME_FORMATS[1]`), and so we should do that here to (since these tests are for the same sessions code).

test/sessions/test_user_session.py
```diff
@@ -29,7 +29,7 @@ class TestUserSession:
         "username": "Test User",
         "name": "Test Session",
         "summary": "Test Summary",
-        "timestamp": datetime.strftime(datetime.now(), DATA_DATETIME_FORMAT),
+        "timestamp": datetime.strftime(datetime.now(), DATA_DATETIME_FORMATS[1]),
```
patrick-austin:
```suggestion
        "timestamp": datetime.strftime(datetime.now(), SESSION_DATETIME_FORMAT),
```
See comment on line 9.

test/sessions/test_user_session.py
```diff
@@ -46,7 +46,7 @@ async def test_valid_get(self):
             "username": "Test User",
             "name": "Test Session",
             "summary": "Test Summary",
-            "timestamp": datetime.strftime(datetime.now(), DATA_DATETIME_FORMAT),
+            "timestamp": datetime.strftime(datetime.now(), DATA_DATETIME_FORMATS[1]),
```
patrick-austin:
```suggestion
            "timestamp": datetime.strftime(datetime.now(), SESSION_DATETIME_FORMAT),
```
See comment on line 9.

test/sessions/test_user_session_list.py
```diff
@@ -3,7 +3,7 @@
 
 import pytest
 
-from operationsgateway_api.src.constants import DATA_DATETIME_FORMAT
+from operationsgateway_api.src.constants import DATA_DATETIME_FORMATS
```
patrick-austin:
```suggestion
from operationsgateway_api.src.constants import SESSION_DATETIME_FORMAT
```
See comments on `test/sessions/test_user_session.py`

test/sessions/test_user_session_list.py
```diff
@@ -25,7 +25,7 @@ async def test_get_session_list(self, _):
                     "summary": "Test Summary",
                     "timestamp": datetime.strftime(
                         datetime.now(),
-                        DATA_DATETIME_FORMAT,
+                        DATA_DATETIME_FORMATS[1],
```
patrick-austin:
```suggestion
                        SESSION_DATETIME_FORMAT,
```
See comments on `test/sessions/test_user_session.py`.

operationsgateway_api/src/records/ingestion/hdf_handler.py
```diff
@@ -108,6 +116,54 @@ async def extract_data(
             self.internal_failed_channel,
         )
 
+    def _parse_data_timestamp(self, timestamp):
```
patrick-austin: Given we have typehints and docstrings for `_create_record_id`, should we have them here too? I guess `timestamp: str | bytes | None` (or maybe just `Any`), and this returns a `datetime`?

operationsgateway_api/src/records/ingestion/hdf_handler.py
```diff
@@ -108,6 +116,54 @@ async def extract_data(
...
+        for fmt in DATA_DATETIME_FORMATS:
+            try:
+                return datetime.strptime(timestamp, fmt)
+            except (ValueError, TypeError) as exc:
+                last_exc = exc
+
+        raise HDFDataExtractionError(
+            "Invalid timestamp metadata. Expected 'timestamp' like "
+            "'2025-04-07T14:28:16.123+00:00' or without milliseconds.",
+        ) from last_exc
```
patrick-austin: Appreciate that before we raised the same message regardless of the root exception being `KeyError, ValueError, TypeError`, so we're no worse off than we were, but do you think it's worth making the messages more specific now we have different `raise` statements? i.e. one for when it's not defined vs when it's defined but can't be parsed.
If not, could also refactor the message so we only define the string once.

operationsgateway_api/src/records/ingestion/channel_checks.py
```diff
@@ -471,6 +490,13 @@ def optional_dtype_checks(self):
                     rejected_channels,
                 )
 
+            if value.metadata.channel_dtype == "string":
```
patrick-austin:
```suggestion
            elif value.metadata.channel_dtype == "string":
```
Very minor efficiency saving, but if line 486 is `True` then 493 can't be True, so we can use `elif` and save the cost of one string equality check.

patrick-austin (changes requested):
A few minor points - the only one I feel somewhat strongly about is using `SESSION_DATETIME_FORMAT` in the tests as well as the equivalent source code being tested for consistency/clarity.

In terms of tests, I think what we want is (let me know if I'm missing anything):
- Reduce overall time taken on CI (by reducing the amount of data ingest needed)
- Explicitly test ingest of all the Gemini features:
  - Sub-second timestampts/record_ids
  - String channels
  - Non-int shotnumbers
- Implicitly test "facility agnostic" existing functionality by running entire suite on ingested "Gemini-style" data

I think we can achieve this with a combination of `pytest.ini` config, fixtures, and `pytest.mark.skipif` in case there's any tests that only make sense in an EPAC or Gemini specific context. I'll probably start this in earnest on Friday, and might need to change approach once I get started - we'll see. 

The main thing I don't have a handle on is real(istic) Gemini data - do you have any existing HDF5 files we could use as the basis for this? Or a script to generate dummy data? I can manually create something which just has the right data to get coverage, but would prefer it to be as realistic as practicable.

test/sessions/test_user_session.py
moonraker595: Agreed, this was just laziness!

operationsgateway_api/src/records/ingestion/hdf_handler.py
moonraker595: Yeah i'm surprised it did moan about duplicated code here. I've refactored it as the same message can be applied to all.

operationsgateway_api/src/records/ingestion/channel_checks.py
moonraker595: Agreed, this was a copy-paste error

moonraker595:
Thanks for this. I'll resubmit for review, but on the understanding that it won't be merged until the tests are in better shape.

I agree with the test plan. `pytest.mark` looks particularly powerful when we want to run the same test against Clara and Gemini but not EPAC, like `@pytest.mark.facilities("clara", "gemini")`

What I'm not sure about is the teardown. Should each facility have a separate ingest process before each set of tests run? Obviously, EPAC will because the data is in a different format, but do we need to do that for Clara & Gemini? (I appreciate there are no tests yet for Clara...just thinking ahead), maybe it could be configurable?

Test data wise, there's plenty of. I have an option to save hdf files to disk from the conversion script (the thing that copies the data to OG), or there's the [HDF generator script](https://github.com/ral-facilities/operationsgateway-hdf5-generator) which allows you to put anything you want in an HDF file, handy if you want invalid ones. Just let me know what you need.

patrick-austin (approved):
Changes look good.

> I'll resubmit for review, but on the understanding that it won't be merged until the tests are in better shape.

That sounds good to me, obviously if there's a functional need to merge this sooner (e.g. Louise needs it on main, or to avoid conflicts with subsequent development) then that's fine too but it would be convenient to keep this open while I work on the tests, if for nothing else then to make it easier to track the lines added that will need coverage.

> What I'm not sure about is the teardown.

I have two ideas that might work. One would be to move the ingest to a fixture which runs before the tests session, then tears down once complete. The other would be to have multiple buckets for each facility, and ingest to those via whatever method. In both cases, I think we'd indicate the facility via the config settings and run `pytest` for each facility. i.e. `pytest --config-file=epac-pytest.ini ...` where `epac-pytest.ini` can then define some/all of the config settings via Pydantic settings sources. I think the latter would be more convenient, as you would not need to re-ingest locally, just replace the current single bucket with two new ones, ingest a smaller amount of data to each, and then run the two test suites as needed.

Things might work out differently of course, but I'm hopeful it could work that way. It would also then be a case of a third bucket for CLARA or other facility specific formatting.

> Just let me know what you need.

Thanks, I'll let you know what I start on this in earnest.

operationsgateway_api/src/config.py
```diff
@@ -112,6 +112,8 @@ class AuthConfig(BaseModel):
     refresh_token_validity_days: StrictInt
     fedid_server_url: StrictStr
     fedid_server_ldap_realm: StrictStr
+    allow_user_office_login: StrictBool
+    user_office_api_key: StrictStr
```
patrick-austin: Is this intended to be a required field? It hasn't been added to `operationsgateway_api/config.yml.example` or `.github/ci_config.yml`, which means that the tests won't run. If `allow_user_office_login` is `False`, is it right to require it to still be set to any string (as currently) or would it be better to make it optional?

operationsgateway_api/src/auth/authentication.py
```diff
@@ -160,3 +161,218 @@ def create_tokens_response(user_model):
...
+            log.info(
+                "User Office login successful for '%s', userId '%s'",
+                username,
+                user_id,
+            )
+
+            return user_id
+
+        except UnauthorisedError:
+            raise
```
patrick-austin: Is the intention to do something here other than just `raise`? If not then should we omit these lines (like how there's no `except ...: raise` for `AuthServerError`).

operationsgateway_api/src/auth/authentication.py
```diff
@@ -160,3 +161,218 @@ def create_tokens_response(user_model):
...
+            if not lookup_response:
+                log.info("No User Office account found for '%s'", email)
+                return None
+
+            user_id = lookup_response[0].get("userNumber")
+
+            if not user_id:
+                log.error(
+                    "User Office lookup response did not contain userNumber"
+                )
```
patrick-austin: `BLK100 Black would make changes.`

operationsgateway_api/src/auth/authentication.py
```diff
@@ -160,3 +161,218 @@ def create_tokens_response(user_model):
...
+                    response.text,
+                )
+                response.raise_for_status()
+                raise AuthServerError()
+
+            lookup_response = response.json()
+
+            if not isinstance(lookup_response, list):
+                log.error(
+                    "Unexpected User Office lookup response format"
```
patrick-austin: `C812 missing trailing comma`

operationsgateway_api/src/auth/authentication.py
```diff
@@ -160,3 +161,218 @@ def create_tokens_response(user_model):
...
+
+            emails = {
+                str(person["userNumber"]): person["email"]
+                for person in lookup_response
+                if (
+                        person.get("userNumber") is not None
+                        and str(person["userNumber"]) in requested_numbers
+                        and "[deactivated]" not in person.get("familyName",
+                                                              "").lower()
+                        and person.get("email")
```
patrick-austin: `E126 continuation line over-indented for hanging indent`
On a related note, would this whole block be more readable if `emails` was initialised to an empty set than added to in a for loop rather than using set comprehension? It seems like there's quite a lot of logic to encapsulate using the comprehension format.

operationsgateway_api/src/auth/authentication.py
```diff
@@ -160,3 +161,218 @@ def create_tokens_response(user_model):
...
+
+            return emails
+
+        except requests.exceptions.RequestException as exc:
+            log.exception("Problem with User Office lookup service")
+            raise AuthServerError() from exc
+
+        except ValueError as exc:
+            log.exception(
+                "Invalid JSON response from User Office lookup service"
```
patrick-austin: `C812 missing trailing comma`

operationsgateway_api/src/auth/authentication.py
```diff
@@ -160,3 +161,218 @@ def create_tokens_response(user_model):
...
+            )
+            raise AuthServerError() from exc
```
patrick-austin: `W292 no newline at end of file`

operationsgateway_api/src/routes/users.py
```diff
@@ -77,6 +78,19 @@ async def add_user(
             )
         login_details.email = email
 
+    if auth_type == "user_office":
+        log.debug("Performing User Office lookup for '%s'",
+                  login_details.username)
```
patrick-austin: `BLK100 Black would make changes.`
Also, if `login_details.username` is a `str` we can use `__repr__` (`%r`) instead of `__str__` (`'%s'`) formatting to get the `''` automatically.

operationsgateway_api/src/routes/users.py
```diff
@@ -77,6 +78,19 @@ async def add_user(
...
+        user_id = Authentication.get_user_id_from_user_office_email(login_details.username)
```
patrick-austin: `B950 line too long (91 > 80 characters)`

patrick-austin (changes requested):
Have looked at the changes from the commits from 67f50f9 onward. Only one functional query about `user_office_api_key` being a required field in the config. The rest is code style stuff, mostly what was flagged by `black`/`flake8` for me locally- I'm assuming the CI didn't run due to the merge conflicts?

operationsgateway_api/src/config.py
moonraker595: I've got rid of the `allow_user_office_login` tag. It'd only be useful if we were planning to allow user office logins and not adding or looking up User Office users. It was there because i had done the user office login work before searching for UO people. Now, the optional `user_office_api_key` acts as the feature switch for User Office functionality.

operationsgateway_api/src/auth/authentication.py
moonraker595: yeah its pointless really, removed

operationsgateway_api/src/auth/authentication.py
moonraker595: done

operationsgateway_api/src/auth/authentication.py
moonraker595: done

operationsgateway_api/src/auth/authentication.py
moonraker595: yeah it made sense at the time just to string the conditions together, I've refactored.

operationsgateway_api/src/auth/authentication.py
moonraker595: done

operationsgateway_api/src/routes/users.py
moonraker595: done

moonraker595:
Thanks for the review. I had a few merge issues, as this branch held the test refactor, but main held the inferred bit-depth functionality. I'll ping it back to you if thats ok, if you could double-check that I haven't blatted over any of your image work.

operationsgateway_api/src/routes/users.py
moonraker595: oh i missed this one, done

operationsgateway_api/src/auth/authentication.py
```diff
@@ -327,8 +323,7 @@ def get_user_office_emails(
                 },
                 headers={
                     "Authorization": (
-                        f"Api-key "
-                        f"{Config.config.auth.user_office_api_key}"
+                        f"Api-key " f"{Config.config.auth.user_office_api_key}"
```
patrick-austin:
```suggestion
                        f"Api-key {Config.config.auth.user_office_api_key}"
```

test/endpoints/test_get_crosshair_intensity.py
patrick-austin: I'm happy with condensing the two `parametrize` blocks down to one, but I think we should keep the tests for both a 16 bit (`FE-204-NSO-P1-CAM-1`) and 12 bit (`CM-202-CVC-CAM-1`) image to protect against regression for https://stfc.atlassian.net/browse/DSEGOG-557 (this was also merged to `main` while this PR was open alongside the inferred depth stuff). It will probably mean refactoring some of the values back out (record id, pixel values, cross hair position).

operationsgateway_api/src/records/export_handler.py
```diff
@@ -586,7 +588,7 @@ def _add_value_to_csv_line(
         if value is None or value == "":
             # leave cell empty in these cases
             return line + ","
-        if isinstance(value, str):
+        if type(value) is str:
```
patrick-austin: Isn't `isinstance` the preferred way of doing type checks?

operationsgateway_api/src/records/export_handler.py
```diff
@@ -509,7 +504,7 @@ def _add_main_csv_file_to_zip(self):
         If other files have been exported and therefore a zip file is being prepared
         for export then add the CSV file to the zip file.
         """
-        if self.export_scalars or self.export_strings:
+        if self.export_scalars:
```
patrick-austin: So I can sort of convince myself that we might not need the `export_scalars` and `export_strings` checks around line 300, if say we know that they won't be in the list of channels unless we specifically asked for them to be projected or something, but I can't see why we now no longer need to check `self.export_strings` here? If either could be added to the CSV file then shouldn't we need to check both flags here?

patrick-austin (changes requested):
Only one or two things that stand out when looking at the changes since my last review, which I've left comments one.

Having said that, I might need to do another review of all the commits since it's a bit tricky to pin down lines which may have changed in the merge, or have had changes reverted during later commits (not going to have time tonight though so will leave this review now).

Now that the tests are passing, the patch coverage is failing against the target of 70%. Were you intending to increase coverage further by adding more tests? I've not looked at the missing lines specifically but e.g. the string export stuff seems to have been added but I can't see any explicit tests of it, with my comment on `export_handler.py` I'm not sure if the export would work when `export_scalars` was False but `export_strings` was True.

operationsgateway_api/src/records/export_handler.py
patrick-austin: Alternatively, if the plan is to have just one flag (`export_scalars`) that controls both, I think  it's got other references elsewhere. Not sure what we should do, but think it should be consistent.

moonraker595:
Thanks
> Were you intending to increase coverage further by adding more tests

yeah I was planning on letting Claude loose on the tests and creating a separate PR for it, so the AI slop was in one PR. But as the coverage is failing now, it's probably worth getting it right in this PR (without the AI slop!). I'll address these tomorrow.... including adding tests

test/endpoints/test_get_crosshair_intensity.py
moonraker595: Agreed, as discussed, reverting test back to main version with added precision for Gemini timestamps.

operationsgateway_api/src/records/export_handler.py
moonraker595: yes, we had changed main to use `isinstance` if I remember correctly so changing this to `isinstance`

operationsgateway_api/src/records/export_handler.py
moonraker595: We should be exporting strings, and so we should be checking for them. In fact, this is functionality I have tested with Louise. This is a merge issue; I have restored my original export code.

moonraker595:
> I'm not sure if the export would work when `export_scalars` was False but `export_strings` was True.

Now testing string channels are skipped when string export is disabled, string CSV exports still work when scalar export is disabled, and existing coverage covers when both options are enabled.

In terms of coverage, I had forgotten that Codecov only logs EPAC tests. Still, there's not a lot of code changes that are EPAC-specific. The main culprit is the new UO auth code.

moonraker595:
I've beefed up the test coverage. Fair warning; the last of the auth ones and all of the user ones were AI generated via co-pilot. I think it's in a much better place now, I'll ping it back over to you.

moonraker595:
oh one thing I haven't actually done yet is test all the export stuff (post-merge) via Postman. I'll do that first thing tomorrow.

moonraker595:
> oh one thing I haven't actually done yet is test all the export stuff (post-merge) via Postman. I'll do that first thing tomorrow.

I've checked this manually, and all works as expected. Just didn't want to rely solely on the automated test is all.

moonraker595:
[ebc0b76](https://github.com/ral-facilities/operationsgateway-api/pull/249/commits/ebc0b76f6241589d019693acca917b8bf41e7c3b): noticed a logging issue in which string and scaler values were falling through the if statement, this meant 1000's of duplicated logging lines. This just makes things more explicit.

operationsgateway_api/src/auth/authentication.py
```diff
@@ -160,3 +161,218 @@ def create_tokens_response(user_model):
...
+        User enters email + password.
+        User Office returns userId
+        The returned userId is used to find the user in Mongo.
+        """
+        username = login_details.username
+        password = login_details.password
+
+        log.debug("Doing User Office auth for '%s'", username)
+
+        login_url = "https://api.facilities.rl.ac.uk/users-service/v2/sessions"
```
patrick-austin: Is it OK to hardcode this? Is it Facility specific or would all possible use cases use the same address? I suppose it's only being used for Gemini currently so no harm in leaving it for now, just wondering if there's any benefit of moving it to a config setting in case it helps re-use.

operationsgateway_api/src/auth/authentication.py
```diff
@@ -160,3 +161,218 @@ def create_tokens_response(user_model):
...
+
+            if response.status_code != 201:
+                log.error(
+                    "Unexpected User Office auth response for '%s': %s %s",
+                    username,
+                    response.status_code,
+                    response.text,
+                )
+                response.raise_for_status()
+                raise AuthServerError()
```
patrick-austin: So I guess the idea here is that `raise_for_status` only catches the 4XX and 5XX codes, but we want to raise for all 1XX, 2XX and 3XX codes (excluding 201)? Is this defensive or is there a specific case where the API might return say 200 and we raise in response?

Not a problem either way, just wanted to ask to make sure I'd understood (and if I haven't then maybe need a comment to explain what we're doing).

operationsgateway_api/src/auth/authentication.py
```diff
@@ -160,3 +161,218 @@ def create_tokens_response(user_model):
...
+        """
+        Look up a User Office user by email address.
+
+        Returns the User Office userId if found, otherwise None.
+        """
+        log.debug("Looking up User Office user '%s'", email)
+
+        lookup_url = (
+            "https://api.facilities.rl.ac.uk/users-service/v2/basic-person-details"
+        )
```
patrick-austin: See comments about hardcoding the url.

operationsgateway_api/src/auth/authentication.py
```diff
@@ -160,3 +161,218 @@ def create_tokens_response(user_model):
...
+
+            if response.status_code != 200:
+                log.error(
+                    "Unexpected User Office lookup response for '%s': %s %s",
+                    email,
+                    response.status_code,
+                    response.text,
+                )
+                response.raise_for_status()
+                raise AuthServerError()
```
patrick-austin: See other comments.

operationsgateway_api/src/auth/authentication.py
patrick-austin: I suppose at least one advantage of doing it this way is that you get the exact code from `raise_for_status` for the traditional failure codes, and only need to use the generic 500 from `AuthServerError` when it doesn't `raise_for_status` first.

operationsgateway_api/src/auth/authentication.py
```diff
@@ -160,3 +161,218 @@ def create_tokens_response(user_model):
...
+
+        Returns a mapping of User Office user numbers to email addresses.
+        Accounts marked as deactivated or without an email address are
+        excluded from the returned mapping.
+        """
+
+        lookup_url = (
+            "https://api.facilities.rl.ac.uk/"
+            "users-service/v2/basic-person-details/search"
+        )
```
patrick-austin: See other comments about hardcoded url.
Even if we don't refactor to config, could refactor the `https://api.facilities.rl.ac.uk/users-service/v2/` bit to a class variable to cut down on repetition/potential sources for error?

operationsgateway_api/src/auth/authentication.py
```diff
@@ -160,3 +161,218 @@ def create_tokens_response(user_model):
...
+            )
+
+            if response.status_code != 200:
+                log.error(
+                    "Unexpected User Office lookup response: %s %s",
+                    response.status_code,
+                    response.text,
+                )
+                response.raise_for_status()
+                raise AuthServerError()
```
patrick-austin: See other comments.

operationsgateway_api/src/records/ingestion/channel_checks.py
```diff
@@ -91,7 +93,7 @@ async def channel_dtype_checks(self):
         """
         ingested_channels = self.ingested_record
         dump = False
-        if not isinstance(ingested_channels, dict):
+        if type(ingested_channels) is not dict:
```
patrick-austin: Did you come to a decision on `isinstance(...)` vs `type(...) is ...`? As I understand it, in principle the former is "preferred" since it allows subclasses, and the latter should only be used for exact type matching. In terms of the API, on this branch we have 68 `isinstance(...)` checks and 4 `type(...) is ...` checks, so looks like (historically at least) we've favoured the former.

operationsgateway_api/src/records/ingestion/channel_checks.py
```diff
@@ -513,7 +539,7 @@ def _waveform_dataset_check(self, rejected_channels, value, key, letter):
         this generates a rejected channel message depending on what was fed to this
         function (which dataset and why it failed)
         """
-        if not isinstance(value, list):
+        if type(value) is not list:
```
patrick-austin: See comments above.

.github/workflows/ci.yml
patrick-austin: Regarding coverage from EPAC vs Gemini modes of operation, pytest has [`--cov-append`](https://pytest-cov.readthedocs.io/en/latest/config.html#cmdoption-cov-append) which sounds good, main challenge I guess would be the fact the GHA run in parallel, so might need to pass artifacts around or have them run sequentially. Alternatively should be possible to run them in parallel and then just call [`coverage combine`](https://coverage.readthedocs.io/en/7.16.1/commands/cmd_combine.html#combining-data-files-coverage-combine) on the outputs in a later job to upload to codecov?

At any rate I think this is probably suitable as a nice to have/follow up issue, and I'll take the codecov report with a pinch of salt for now.

test/endpoints/test_get_crosshair_intensity.py
```diff
@@ -60,11 +62,18 @@ def test_get_crosshair_intensity(
...
-            url = f"/images/{record_id}/a/crosshair?functions={quote(functions)}"
+            row = [value / 10 for value in row]
+            column = [value / 10 for value in column]
+            functions = json.dumps(
+                {
+                    "name": "a",
+                    "expression": f"{channel_name} / 10",
+                },
+            )
+            url = f"/images/{record_id}/a/crosshair" f"?functions={quote(functions)}"
```
patrick-austin:
```suggestion
            url = f"/images/{record_id}/a/crosshair?functions={quote(functions)}"
```

patrick-austin (changes requested):
Had a look over it again and could only find one (very) minor formatting change I'd definitely make (`test_get_crosshair_intensity.py`).

The rest are things that I noticed and wanted to check you're happy with, but in principle I don't think  necessarily need to change if you are. As far as I could see the codecov report is OK given that we won't have the Gemini suite contributing to the coverage (I didn't consider on my last review until you pointed it out). I've left some thoughts about that as well but think it would be fine to let it lie (either for this PR and pick up as a separate issue, or for the foreseeable future).

One final thing - I was asking Kevin what bits of the API code would be covered by the "data integrity" testing he's been doing and if he knew what version/commit the Gemini instance is running, since I wasn't sure if any changes made to this branch would affect things for him. I think the answer was essentially "no" since he wasn't using export, and I guess at some point we'd have to deploy the latest version of the API to the Gemini instance anyway so it would get tested in pre-prod whenever that happens.

operationsgateway_api/src/records/ingestion/channel_checks.py
moonraker595: Agreed, i think I originally changed it because a newer version of the linter was moaning. I've reverted all type(...)  to isinstance(...)

.github/workflows/ci.yml
moonraker595: have created ticket: https://stfc.atlassian.net/browse/DSEGOG-571 to capture this

operationsgateway_api/src/auth/authentication.py
moonraker595: > Is it OK to hardcode this?

Not really, This was corner-cutting on my part. I've refactored this to sit behind a global flag. It was getting too complicated to switch functionality behind the API key or URL, so I added the global flag back in, with a model validator to check for the key and URL if the flag is set. The other deployments just need to set this to false.

I've noticed that there is also: `experiments/user_office_rest_api_url`, which is slightly different but very similar. I'm happy with these both being in the config.

operationsgateway_api/src/auth/authentication.py
moonraker595: I don't think the raise_for_status() call is strictly necessary here since we already log the response status and body, and we're not using the extra info in the testing anyway. This is almost certainly a response to some debugging where I wasn't happy with the response being spewed out for some reason. I'll remove it for now since we've got the log.error and i can't see a need to keep both.

moonraker595:
Thanks, have made the appropriate changes.

With regards to Kevin's testing; As you mentioned, he is only testing the data, not the version/functionality of OG. 
The version deployed to prod (there is no pre-prod) is commit `09f295d58d7e270758650a3556acf28d903ad115` which is from July 23rd and holds most of the UO functionality used for testing with Louise. Once this is commited into main i need to come up with a release plan which will entail:
1st: Gemini and test
2nd: Epac pre-prod and test, then prod
3rd: Clara, using it as a new base for pre-prod and prod (there's nothing on prod at the mo, so it needs setting up)

operationsgateway_api/src/config.py
```diff
@@ -112,9 +113,31 @@ class AuthConfig(BaseModel):
...
+                for name in ("user_office_api_key", "user_office_users_service_url")
+                if not getattr(self, name)
+            ]
+            if missing:
+                raise ValueError(
+                    f"{', '.join(missing)} must be set when "
+                    "'user_office_integration' is enabled",
+                )
+
+        return self
```
patrick-austin: An alternative way of doing this could be to define an optional `UserOfficeConfig` that is optional in the main config, but with both the API key and url being required within the model itself. That would mean we don't need custom validation since Pydantic will allow nothing to be provided (i.e. `user_office_config=None`) or both, but if only one is provided then `UserOfficeConfig is missing a required field and it will fail.

patrick-austin (commented):
Changes look good - just had one comment about the custom validator for the user office config. I think it's fine as is but was wondering if an optional "sub-model" would let us get the same functionality without having to write/test a custom validator? If there's a reason I'm missing why that wouldn't work, or you've already considered it and prefer it this way then I'm happy to approve, but thought it was worth raising for consideration at least.

moonraker595:
I initially assumed a sub-model would require moving the User Office settings out of the auth section, but I suppose it can be an optional nested model within AuthConfig, so a sub sub model. So the config would look like:
```
auth:
  refresh_token_validity_days: ...
  fedid_server_url: ....
  fedid_server_ldap_realm: ....

  user_office:
    api_key: ...
    users_service_url: ...
```
So the user office section is optional, and other deployments just won't have. Sound ok?

operationsgateway_api/src/config.py
moonraker595: Done. There are a few files changed, but they're mostly single-line changes. I had forgotten we could check on the empty tag, and we do this elsewhere, so this is a lot neater. Good catch.

patrick-austin (approved):

PR #252, opened 2026-05-07
Overhaul tests to account for Gemini data format and features

patrick-austin:
## CI
* Add minimal `permissions` to `ci.yml` (unrelated to main changes, but best practice)
* Replace single `tests` job in `ci.yml` with two jobs for epac/gemini so they can run simultaneously 
* Use environment variables to override config settings rather than modifying in place with yq (see changes to config.py)
* Comment out settings in `.github/ci_config.yml` and `.github/ci_ingest_echo_config.yml` which are intended to be overridden by environment variables

## Source code changes
* Except `"string_type"` `ValidationError`s instead of numerics in `hdf_handler.py`
  * Having written the test for this block, when you set data to e.g. `0` rather than a string, it throws `"string_type"` not `"int_type"`. 
* Replace `StrictBool` with `bool` for `use_sub_second_timestamps`
  * Unfortunately I couldn't get `StrictBool` to work with the environment variables since I think these will always come through as a `str` and have to be coerced to `bool`
* Replace the custom `load` function with the Pydantic native `settings_customise_sources` to allow environment variable overrides
* Some quality of life refactors to the ingest code to make it easier to work with

## Tests
* Move `reset_record_storage` to main `conftest.py` as it's used in `test/records` as well as `test/endpoints`, so it should be visible to both automatically without need for explicit import
* Refactor most instances of record_ids, timestamps in the tests to global variables defined in `conftest.py`. These are formatted appropriately based on `Config.config.app.use_sub_second_timestamps`
  * In some cases, this is sufficient to make the tests "facility agnostic", i.e. the test will pass without further changes to the inputs or asserts
  * In others, the inputs or asserted values are different, so in these cases the global record_ids don't need to be used since those tests can't run for both facilities anyway
* Account for the removal of most of the test hdf files:
  * Reducing record counts
  * Changing dates/record ids to be one of the 4 that remain
  * Changing data values to match one of those 4
* For tests which are explicitly tied to EPAC or Gemini formatting (i.e. those using shotnum), use `pytest.mark` inside `pytest.param` to prevent them from running in the wrong mode, and add Gemini equivalents for the EPAC tests.
* Some non-functional test refactors (removing parameterized inputs that don't actually vary)
* Add string channel to manifest
* Add explicit Gemini format tests to `test_range_converter.py` to cover new functionality
* Rename/modify/remove/add files to `test/export/` to reflect removal of certain records, and Gemini versions of shotnums/timestamps.
  * NOTE: This accounts for +1500/-6000 of the lines changed, as each waveform CSV has ~1500 lines
* Add string channel to `create_test_hdf_file` alongside the other channel types, and use this for explicit tests of ingest functionality (in particular error handling)
* Add fixture to run the ingest before tests iff no records present
  * Currently, there isn't much difference here between running the ingest explicitly with the script vs relying on this fixture. For me, the main motivations for adding this are:
    * When running tests locally, the same environment variables (i.e. via `pytest_gemini.ini` or `pytest_epac.ini`) will apply to this fixture as to the actual tests, which consolidates down to less sources of truth for config.
    * In principle, you could add tear down after the yield. I don't think we need this currently (since keeping everything in your local DB/storage makes local testing much quicker than a full tear down) but it's nice to have the option.
    * You could also parametrize this fixture if you wanted to run both EPAC and Gemini ingest/tests back to back. I don't think we want to do this (since running them in parallel halves CI time compared to in series) but again it leaves the option open.
* Remove test of redundant `Config.load`

## Documentation
* Add section to `docs/developer/test_data.md`
* Add sections to `README.md`


test/pytest.ini.example
```diff
@@ -0,0 +1,20 @@
+[pytest]
+asyncio_mode=auto
+env = 
+    ; EPAC ingest config overrides
+    ; DATABASE__CONNECTION_URI=mongodb://localhost:27017/opsgateway-epac
+    ; SOURCE__SIMULATED_DATA_BUCKET=og-ci-simulated-data-epac
+    ; API__STORAGE_BUCKET=og-patrick-dev-epac
```
moonraker595:
```suggestion
    ; API__STORAGE_BUCKET=og-yourname-dev-epac
```

moonraker595 (changes requested):
Overall it looks really good and seems to have cut down ingestion on the CI by half.

Locally:

- Epac: Total ingestion duration: 283.74 seconds
- Gemini: Total ingestion duration: 291.11 seconds

Which is definitely quicker.

Further to our discussion, I cleared all the echo buckets and dbs locally and rerun the pytest inis without any issues. I would maybe put the pytest ini's in their own config and check those in uncommented.

Overall, great work, and thanks for helping with this.

patrick-austin:
> I would maybe put the pytest ini's in their own config and check those in uncommented.

Agreed. Latest commit replaces `pytest.ini.example` with two facility specific `.ini.example` files with the variables uncommented. Also updated the README to reflect this change, and give a bit more explanation of how pytest/pydantic settings and overrides work.

moonraker595 (approved):
👍

PR #253, opened 2026-06-09
Checking that refresh token now contains the username

moonraker595:
Checking that refresh token now contains the username and it matches the corresponding access token. Nothing changes on the FE.

To recap, the flow is now:

on `/login`

- API authenticates users in existing ways
- API returns an access JWT in the response body.
- API also sets a refresh JWT as an HttpOnly cookie.
- Access token contains:
    - username
    - authorised_routes
    - userIsAdmin
    - exp
- Refresh token now contains:
    - username
    - exp

Then on /refresh:
- Client sends the existing access token in the body.
- Browser also sends the refresh token cookie.
- API verifies the refresh token signature and expiry.
- API checks the refresh token is not blacklisted.
- API decodes the access token
- API compares username
    - If they match, it extends the access token expiry and returns a new access token.
    - If they do not match, it rejects the refresh.

### DIFF
--- a/operationsgateway_api/src/auth/jwt_handler.py
+++ b/operationsgateway_api/src/auth/jwt_handler.py
@@ -106,6 +106,7 @@ class JwtHandler:
         try:
             access_payload = JwtHandler.get_payload(access_token, {"verify_exp": False})
 
+            # make sure the user associated with the refresh token is also associated with the access token
             if refresh_payload.get("username") != access_payload.get("username"):
                 message = "Refresh token does not match access token user"
                 log.warning(message)
--- a/operationsgateway_api/src/auth/jwt_handler.py
+++ b/operationsgateway_api/src/auth/jwt_handler.py
@@ -66,6 +66,7 @@ class JwtHandler:
         :return: The refresh JWT
         """
         payload = {}
+        payload["username"] = self.user_model.username
         payload["exp"] = datetime.now(timezone.utc) + timedelta(
             days=Config.config.auth.refresh_token_validity_days,
         )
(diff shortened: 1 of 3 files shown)


operationsgateway_api/src/auth/jwt_handler.py
```diff
@@ -95,18 +96,30 @@ def refresh_token(refresh_token: str, access_token: str):
...
         if refresh_token in JwtHandler.get_blacklisted_tokens():
             message = "Token is blacklisted"
             log.warning(message)
             raise ForbiddenError(message)
         try:
-            payload = JwtHandler.get_payload(access_token, {"verify_exp": False})
-            payload["exp"] = datetime.now(timezone.utc) + timedelta(
+            access_payload = JwtHandler.get_payload(access_token, {"verify_exp": False})
+
+            # make sure the user associated with the refresh token is also associated with the access token
```
patrick-austin: Think this is the line causing a linting failure

patrick-austin (commented):
Changes look fine to me, but I think the linter is complaining about the comment being too long. I'm assuming that all the safety failures would be fixed by the PR that bumped to mongo8? Could merge in main/rebase to verify.

operationsgateway_api/src/auth/jwt_handler.py
moonraker595: thanks, i'll address this and resubmit. I think some more have come about since I added them to the mongo8 branch. I'm going to address any safety issues by adding them to the dependabot branch.

patrick-austin (approved):

